PR #56, opened 2026-05-11
feat: bake workflow extensions into core

TheLonelyDevil9:
## What?
Bakes three workflow extensions into SillyBunny core extensions:

- Adds a trimmed `guided-generations` core extension with Simple Send, Guided Response, Guided Swipe, Guided Impersonate, and QR-bar integration.
- Adds `input-history` as a core extension for input recovery/history.
- Adds `quick-image-gen` as a core extension using the existing Quick Image Gen runtime bundle.
- Removes these extensions from the optional starter-pack/Launchpad install list and removes the old Guided Generations fork warning.

## Why?
These extensions should ship as part of the SillyBunny distribution instead of depending on separate user-installed extension repositories. That avoids duplicate installs, fork warnings, and starter-pack entries for features that are now bundled.

Input recovery should have one owner. The bundled Input History extension is the recovery/history implementation, so Guided Generations no longer ships a separate recovery button.

## How?
Guided Generations was reduced to the requested subset and moved under `public/scripts/extensions/guided-generations` with a new `guided-generations` settings key plus one-time migration from the old `GuidedGenerations-Extension` key. The old Guided Generations recovery setting is dropped during settings load so stale local settings do not keep a dead option around.

Input History was vendored under `public/scripts/extensions/input-history` with runtime import timing adjusted for core loading.

Quick Image Gen was vendored under `public/scripts/extensions/quick-image-gen`, with its dynamic import paths adjusted for the core-extension location. Its settings key remains `quick-image-gen`, so existing QIG settings continue to apply. The QIG runtime bundle is lint-ignored as vendored code to preserve upstream formatting while still validating it with syntax and browser-load checks.

## Testing?
- `node --check public/scripts/extensions/quick-image-gen/index.js`
- `node --check public/scripts/extensions/guided-generations/index.js`
- `node --check public/scripts/extensions/guided-generations/scripts/*.js`
- `node --check public/scripts/extensions/input-history/index.js`
- `node --check public/scripts/extensions/input-history/lib/wait.js`
- `node --check public/scripts/extensions.js`
- `node --check public/scripts/welcome-screen.js`
- `git diff --check`
- `eslint --resolve-plugins-relative-to <installed repo deps> "src/**/*.js" "public/**/*.js" "scripts/**/*.js" ./*.js`
- Browser smoke test against a running SillyBunny instance: extension discovery reported `guided-generations`, `input-history`, and `quick-image-gen` as `system`; no old third-party IDs for those extensions remained; the browser console had no errors; QIG manifest/script/style loaded successfully.

## Screenshots (optional)
No screenshots. This is primarily extension packaging/runtime loading work; browser verification checked discovery, network loads, and console health instead.

## Anything Else?
The local user-installed copies were removed from the runtime checkout, but `data/default-user` is intentionally not part of this PR.


TheLonelyDevil9:
> Removes these extensions from the optional starter-pack/Launchpad install list and removes the old Guided Generations fork warning.

@platberlitz just check if you're okay with that, and if any polish is needed with Launchpad in general. I say let's clean house there too and remove whatever we currently agree on (older view was different I say)

platberlitz:
> > Removes these extensions from the optional starter-pack/Launchpad install list and removes the old Guided Generations fork warning.
> 
> @platberlitz just check if you're okay with that, and if any polish is needed with Launchpad in general. I say let's clean house there too and remove whatever we currently agree on (older view was different I say)

I agree with deleting them from the Launchpad, yeah. It's pointless since it'd just duplicate them.
What do we do when users have installed the extensions already? We can make a semi-permanent toast to warn them about it and add a button to easy-disable or uninstall them?

TheLonelyDevil9:
> We can make a semi-permanent toast to warn them about it and add a button to easy-disable or uninstall them?

Pretty much this, I believe there was a toast already that I saw on staging. We can make it clearer they need to get rid of the extension since it's baked in

PR #57, opened 2026-05-12
fix: confirm preset overwrites and unsaved text edits

TheLonelyDevil9:
## What?
Adds confirmation coverage to preset saves and unsaved preset text edits:

- Confirms before overwriting an existing preset or advanced-formatting template.
- Tracks textareas and text inputs associated with preset/template managers.
- Warns before switching presets/templates when those text fields have unsaved edits.
- Adds a browser `beforeunload` warning while tracked text edits are dirty.

## Why?
Preset text fields can hold large prompt/template edits. Before this change, switching presets or leaving the page could silently discard those edits, and save/update could overwrite the selected preset without a final confirmation.

## How?
`PresetManager` now snapshots associated preset text fields, listens for text input changes, and installs a capture-phase select guard that reverts the select until the user confirms discarding edits. Confirmed switches dispatch a normal DOM change event so existing preset handlers still apply the target preset UI. Saves, preset loads, and preset-change events refresh the snapshot baseline. OpenAI non-UI preset saves emit the same preset-change event so dirty tracking clears after successful update-button saves.

## Testing?
- `npm run lint -- --quiet`
- `git diff --check`
- `git diff --check HEAD~1..HEAD`
- Raw `npm run check:frontend-budgets` in the Windows worktree reports blocking stylesheets at 692.1 KiB over the 690.0 KiB budget; this patch only changes JS and the edited files are LF in the worktree.
- LF committed-tree archive budget check passed: blocking stylesheets 14 / 672.6 KiB, startup scripts 21 / 1547.8 KiB.
- Browser smoke on local Node server: edit system prompt text, switch template, unsaved dialog appears, cancel keeps original selection and dirty text, accept switches to the target template, repaints the textarea, and clears dirty/beforeunload state.

## Screenshots (optional)
Not included. The flow uses the existing popup component and was verified through targeted browser DOM state checks.

## Anything Else?
This stays scoped to preset-manager text fields (`textarea` and `input[type="text"]`). It does not broaden dirty tracking to sliders, checkboxes, or unrelated settings controls.

### DIFF
--- a/public/scripts/openai.js
+++ b/public/scripts/openai.js
@@ -6324,6 +6324,10 @@ async function saveOpenAIPreset(name, settings, triggerUi = true) {
             option.innerText = data.name;
             if (triggerUi) $('#settings_preset_openai').append(option).trigger('change');
         }
+
+        if (!triggerUi) {
+            await eventSource.emit(event_types.PRESET_CHANGED, { apiId: 'openai', name: data.name });
+        }
     } else {
         toastr.error(t`Failed to save preset`);
         throw new Error('Failed to save preset');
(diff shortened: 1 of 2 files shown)


platberlitz:
Seems good to me.

PR #64, opened 2026-05-12
fix: restore avatar preview and shape styling

TheLonelyDevil9:
## What?
Restores the avatar shape rules that are still relevant after the character menu rework: square avatars stay square, big avatars keep their 3:4 proportions, avatar images inherit their parent shape, and zoomed avatar overlays sit above the SillyBunny modal/content layers.

## Why?
Avatar shape modes could visually collapse back toward round defaults because image-level border radii did not reliably inherit the selected avatar mode. Zoomed avatars could also sit under newer shell/modal layers.

## How?
Keeps the fix scoped to existing shape-mode CSS and overlay stacking. The older character-editor preview visibility block was removed after the new character card menu added scoped editor preview sizing.

## Testing?
- `git diff --check`
- `node scripts/check-frontend-budgets.js`
- `node --check server.js`

## Screenshots (optional)
Not included; this is a small CSS cleanup rebased after the character menu rewrite. Browser smoke was skipped to keep the runtime checkout and default `localhost:4444` server untouched while the isolated dev worktree is being finalized.

## Anything Else?
Reworked from the original draft per review feedback after the new character card menu landed. The PR no longer changes editor preview visibility/sizing directly.

### DIFF
--- a/public/css/sillybunny-theme.css
+++ b/public/css/sillybunny-theme.css
@@ -130,6 +130,13 @@
     --sb-message-radius: 22px;
 }
 
+.zoomed_avatar_container,
+.avatar_zoom_container,
+.zoomed_avatar,
+.avatar_zoom {
+    z-index: calc(var(--sb-z-modal-content) + 5) !important;
+}
+
 :root[data-sb-compact-mode='true'] {
     --sb-control-min-height: 32px;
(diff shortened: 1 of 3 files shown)


Geechan:
@TheLonelyDevil9 Draft this until the new character card menu is implemented. Will likely need to be reworked.

TheLonelyDevil9:
Reworked

PR #66, opened 2026-05-12
feat: expand connection profile summaries

TheLonelyDevil9:
## What?
Expands Connection Profiles saved summaries with active agent, sampler, template, system prompt, and World Info counts, then renders the summary in a two-column layout.

## Why?
Connection profile saves currently capture core command state but leave out important contextual settings that users need to compare before applying a profile.

## How?
Adds a snapshot enrichment pass after profile command reads, exposes enabled in-chat agents through a small global bridge, collects sampler/template/system prompt/world-info values from current UI/context, and styles the profile summary list with a two-column grid.

## Testing?
- `npm run lint -- --quiet`
- `git diff --check`

## Screenshots (optional)
Not included; no browser server was started for this split.

## Anything Else?
Values are summary metadata only; profile application remains command-driven and backwards-compatible.

### DIFF
--- a/public/scripts/extensions/connection-manager/index.js
+++ b/public/scripts/extensions/connection-manager/index.js
@@ -89,6 +89,12 @@ const FANCY_NAMES = {
     'prompt-post-processing': 'Prompt Post-Processing',
     'secret-id': 'Secret',
     'regex-preset': 'Regex Preset',
+    'active-agents': 'Active Agents',
+    'samplers': 'Samplers',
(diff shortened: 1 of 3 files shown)


TheLonelyDevil9:
@Geechan @platberlitz let me know what else connection profiles should capture, happy to modify this one

platberlitz:
Can't really think of any, that seems comprehensive enough.

TheLonelyDevil9:
129 LOC, good to merge?

PR #67, opened 2026-05-12
feat: add OOC and HTML context depth controls

TheLonelyDevil9:
## What?
Adds user-configurable OOC and raw HTML context-depth controls for prompt assembly.

## Why?
The current prompt builders strip all user-visible OOC blocks and raw HTML tags from context, but users may want the most recent N messages to preserve those details while older context stays clean.

## How?
Adds `ooc_context_depth` and `html_context_depth` power-user settings, UI number inputs beside tag handling, preservation-aware OOC/HTML stripping helpers, and applies the depth checks in both text-completion and chat-completion prompt assembly.

## Testing?
- `npm run test:unit --prefix tests -- tests/ooc-blocks.test.js`
- `npm run lint -- --quiet`
- `git diff --check`

## Screenshots (optional)
Not included; settings UI is a small form-control addition and no browser server was started.

## Anything Else?
Default depth is `0`, preserving existing behavior for current users.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5776,6 +5776,18 @@
                                     <input id="encode_tags" type="checkbox" />
                                     <small data-i18n="Show tags in responses">Show &lt;tags&gt; in responses</small>
                                 </label>
+                                <div class="flex-container alignitemscenter gap10">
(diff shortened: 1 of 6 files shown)


platberlitz:
Seems good to me.

PR #60, opened 2026-05-12
feat: add pre-generation agent interceptors

platberlitz:
## Summary
- Add pre-generation intercept mode so in-chat agents can modify assembled text or chat-completion context before the main model receives it.
- Add editor settings, schema normalization, docs, and tests for replace, wrap, and patch application modes.
- Bundle Sheep's NPC Motivator as a disabled-by-default pre-generation intercept patch agent.

## Tests
- npm run test:unit --prefix tests -- tests/in-chat-agents-runner.test.js tests/in-chat-agents-store.test.js
- npm run test:unit --prefix tests -- tests/in-chat-agents-templates.test.js tests/in-chat-agents-store.test.js tests/in-chat-agents-runner.test.js
- npm run test:unit --prefix tests -- tests/in-chat-agents-templates.test.js

PR #68, opened 2026-05-12
fix: polish in-chat agent rewrite metadata

TheLonelyDevil9:
## What?
Polishes in-chat agent rewrite metadata, token accounting, undo/redo async handling, and duplicate template additions.

## Why?
Agent rewrites could leave message token counts stale and the transform UI did not expose enough run context. Adding identical agents from templates also had no confirmation step.

## How?
Refreshes message and active-swipe token counts after prompt-transform mutations through the existing extension context, records order/profile/model labels on prompt-transform run metadata, shows those labels in toasts/history rows, updates undo/redo handlers for async recounts, and prompts before adding an identical agent template.

## Testing?
- `npm run test:unit --prefix tests -- tests/in-chat-agents-runner.test.js tests/in-chat-agents-store.test.js`
- `npm run lint -- --quiet`
- `git diff --check`

## Screenshots (optional)
Not included; no browser server was started for this split.

## Anything Else?
This PR intentionally avoids touching the global chat badge code from PR #60, keeping the split focused on the in-chat-agents extension runtime.

### DIFF
--- a/public/scripts/extensions/in-chat-agents/agent-runner.js
+++ b/public/scripts/extensions/in-chat-agents/agent-runner.js
@@ -1435,13 +1435,36 @@ function describePromptTransformTarget(profileId = '', runner = '') {
     return 'the main model';
 }
 
+function getPromptTransformProfileLabel(profileId = '') {
+    return profileId ? getConnectionProfileDisplayName(profileId) : 'Main model';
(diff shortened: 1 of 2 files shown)


platberlitz:
Good QoL.

PR #73, opened 2026-05-12
feat: rework character menu

Geechan:
This is an early implementation of a completely reworked character menu - moving from the old legacy system found in SillyTavern to an overhauled, customised version.

This aligns it much more closely with the rest of the graphical shell design by adding tabs for Character, Group, Persona, and Import functionalities, while overhauling a lot of the confusing UI in the process.

With this: all shell UI panels should be congruent with each other.

Caveats: this has not been tested for mobile, only for desktop, and there will likely remain some gremlins and scaling issues with page size changes, which can be addressed in the future.


TheLonelyDevil9:
Knowing the pain that is UI, great job. I'm good to merge, merging it.

PR #77, opened 2026-05-13
fix: apply geechan review follow-ups

TheLonelyDevil9:
## What?
Applies the pending Geechan review follow-ups for duplicate in-chat agent copy and OOC/HTML prompt-context retention behavior.

## Why?
The duplicate-agent confirmation repeated "Add anyway" in the body and button, and the OOC/HTML context-depth settings needed an explicit default policy. The confirmed SB behavior is now: `-1` keeps all messages in prompt context, `0` strips all OOC/HTML from prompt context, and positive values keep only the newest N context messages.

## How?
Renames the duplicate-agent confirmation OK button to `Add Agent`. Changes OOC and HTML context-depth defaults to `-1`, allows the settings inputs to accept `-1`, adds visible helper copy for the `-1 / 0 / N` semantics, and routes both standard and Chat Completions prompt assembly through a shared retention-depth helper so the two paths stay aligned.

## Testing?
- `npm run test:unit --prefix tests -- tests/ooc-blocks.test.js`
- `npm run lint`
- `git diff --check`

## Screenshots (optional)
Not included; this is settings copy plus prompt-context behavior wiring.

## Anything Else?
This is intentionally split from the Guided Generations fork-toast fix and the broader token-count refresh bug.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5759,16 +5759,18 @@
                                     <small data-i18n="Show tags in responses">Show &lt;tags&gt; in responses</small>
                                 </label>
                                 <div class="flex-container alignitemscenter gap10">
-                                    <label for="ooc_context_depth" class="flex1" title="Keep OOC ((...)) blocks in the last N context messages. 0 strips all OOC blocks from prompt context." data-i18n="[title]Keep OOC blocks in the last N context messages. 0 strips all OOC blocks from prompt context.">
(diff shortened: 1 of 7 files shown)


platberlitz (commented):
Is the OOC stripping thing off by default? It should be, or people are going to get confused and start complaining.

Geechan:
This doesn't actually address the review follow up for OOC default behaviour. It is still enabled by default as I don't see any relevant code changes here.

TheLonelyDevil9:
Updating -1 as `retain all`, 0 as no retention in context, >1 = how many messages of retention in context (intended behavior as per Sheep/Kris)
It whiffed my instructions

Similar changes to HTML retention settings too

TheLonelyDevil9:
default will be -1

platberlitz:
Okay, works for me

platberlitz (commented):
Find this good to go, since it also explains how to use it.

Geechan (approved):
Looks good to me, can merge.

PR #78, opened 2026-05-13
fix: persist guided generations fork warning

TheLonelyDevil9:
## What?
Fixes the Guided Generations legacy-fork warning so it only appears when a separate `third-party/GuidedGenerations-Extension` fork is still installed, and makes the warning sticky until the user closes it with the X button.

## Why?
Users who already uninstalled the old fork could still see the toast because stale migrated settings were treated as proof that the fork was installed. The warning also timed out automatically, so users could miss the conflict notice and keep wondering why Guided Generations behavior was broken.

## How?
Moves the legacy-fork warning decision into a small helper, keys the warning off discovered extension names instead of old settings, switches dismissal persistence from session storage to local storage, and only writes the dismissal flag from the toast close-button handler. The toast now uses `timeOut: 0`, `tapToDismiss: false`, and `closeButton: true`.

## Testing?
- `npm run test:unit --prefix tests -- tests/guided-generations-toast.test.js`
- `npm run lint`
- `git diff --check`

## Screenshots (optional)
Not included; behavior is covered by the focused unit test and follows the existing sticky-toast pattern used elsewhere in the app.

## Anything Else?
This PR is intentionally split from the Geechan review follow-up and the broader token-count refresh bug.

### DIFF
--- a/public/scripts/extensions/guided-generations/index.js
+++ b/public/scripts/extensions/guided-generations/index.js
@@ -4,11 +4,14 @@ import { extensionName, getPresetsForApiType, getProfileApiType, getProfileList 
 import { guidedImpersonate } from './scripts/guidedImpersonate.js';
 import { guidedResponse } from './scripts/guidedResponse.js';
 import { guidedSwipe } from './scripts/guidedSwipe.js';
+import {
+    markOldExtensionWarningDismissed,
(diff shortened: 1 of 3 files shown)


platberlitz:
Good to go.

PR #81, opened 2026-05-13
fix: refresh message token counts after edits

TheLonelyDevil9:
## What?
Refreshes saved message token metadata after message text or reasoning changes outside the normal generation path.

This covers normal message edit completion, in-chat agent prompt transforms such as Prose Polisher, translated outgoing messages, image-caption replacements for empty messages, and manual reasoning edit/delete actions. Active swipe metadata now receives the same refreshed `token_count` and `reasoning_tokens` as the visible message.

## Why?
The previous fix only refreshed counts in the in-chat agent rewrite path, leaving other UI-driven message mutations able to save stale `extra.token_count` values. That meant tools like Prose Polisher could visibly change a message while the token count stayed at the old value, and swipe metadata could remain stale even when the current message metadata was refreshed.

## How?
Routes prompt-transform recounts through the shared message token accounting helper instead of a private mini-counter, updates that helper to mirror refreshed output/reasoning counts into the active swipe, and calls the shared helper from the direct text/reasoning mutation paths that save or render changed content.

The change stays backward compatible with existing chats: it only updates existing `extra` metadata when a message is touched, and it leaves inactive swipe metadata unchanged.

## Testing?
- `npm run test:unit --prefix tests -- tests/reasoning-token-accounting.test.js`
- `npm run test:unit --prefix tests -- tests/in-chat-agents-runner.test.js`
- `npm run lint`
- `git diff --check`

## Screenshots (optional)
Not included; this is metadata/accounting behavior covered by unit tests rather than a visual UI change.

## Anything Else?
Root dependencies were installed in the worktree to run lint; npm reported pre-existing audit findings during install, but no dependency files were changed.


platberlitz:
I tried testing it in both mobile and desktop and it doesn't seem to be working for me. Double-checked the branch and did a cache + cookie refresh...

Used Prose Polisher multiple times, and edited the message myself for testing. Didn't do anything.

TheLonelyDevil9:
Testing again to ensure functionality

TheLonelyDevil9:
What changed:
- `updateMessageMetaBadges()` now also syncs `.tokenCounterDisplay` from `message.extra.token_count`, so Prose Polisher, translation, reasoning edits, and other `updateMessageBlock()` paths update the visible badge.
- Manual message edit now calls the same badge updater after recalculating token accounting.
- Added `formatTokenCounterText()` so empty/invalid/zero counts clear the badge instead of leaving stale text.

Pushed commit: `58ed9d6f6 fix: sync visible token counters after edits`

Validation:
- `npm run lint` passed
- `npm run test:unit --prefix tests` passed, 427 tests
- `npm run check:frontend-budgets` passed
- GitHub PR checks passed

PR #82, opened 2026-05-13
Fix input history controls in chat composer

platberlitz:
## Summary
- Move input history controls into the top composer action row, sharing the Guided Generations row when it exists and falling back to a standalone row above the textarea.
- Restyle the controls to match the Guided Generations icon sizing, including compact 28-34px mobile buttons with 8px radius.
- Flatten the history chevrons into a row and make the history menu open downward from the controls.

## Verification
- npm run lint initially failed because the environment resolved global ESLint 10 while the repo uses .eslintrc.cjs.
- npm install
- npm run lint -- --quiet
- git diff --check

### DIFF
--- a/public/scripts/extensions/input-history/index.js
+++ b/public/scripts/extensions/input-history/index.js
@@ -29,6 +29,37 @@ let arrowsWrap;
 let btnHistory;
 /**@type {HTMLElement} */
 let historyMenu;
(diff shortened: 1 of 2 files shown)


TheLonelyDevil9:
Good idea, I was still trying to balance left and right

PR #83, opened 2026-05-13
feat(reasoning): add xhigh effort, rename auto to blank

TheLonelyDevil9:
## What?

Adds `xhigh` reasoning effort level and renames `auto` to `blank` across the reasoning effort system.

- `blank` replaces `auto` in the UI select, `reasoning_effort_types`, default settings, and all budget switch cases. Semantics are identical: do not send the field.
- `xhigh` is added as a new selectable level. It passes through to providers that support it natively.
- `max` now resolves to `xhigh` (instead of `high`) for OpenAI models `gpt-5.2+` and xAI `grok-4.20-multi-agent`. For all other providers/models it still resolves to `high`.
- Adaptive Claude (`xhigh`) maps to `'max'` effort string. Traditional Claude (`xhigh`) uses the same 95% budget as `max`.
- All Google budget functions handle `xhigh` (same ceiling as `max`) and `blank` (same as `auto`).
- XAI frontend filter updated: `grok-4.20-multi-agent` is now allowed to send `reasoning_effort` alongside `grok-3-mini`.
- XAI backend handler updated: passes `xhigh` through instead of clamping everything to `high`/`low`.
- Migration: saved settings containing `'auto'` are normalized to `'blank'` on load via `migrateChatCompletionSettings`.

## Why?

OpenAI introduced `xhigh` for models after GPT-5.1 Codex Max. xAI added it for `grok-4.20-multi-agent`. The existing `auto` label was misleading — it never sent a value, it just omitted the field. `blank` is more accurate. `max` as an adaptive alias should now reach the highest available effort level rather than capping at `high`.

## How?

- `xhigh` passes through the payload unchanged for OpenAI/Responses/Azure/AIMLAPI/OpenRouter/Pollinations/Perplexity/CometAPI/ElectronHub/Chutes — no mapping needed.
- The `max` → `xhigh` promotion in `resolveReasoningEffort` is gated on a model regex (`/^gpt-5\.(2|3|4|5)/`) so older GPT-5 and o-series models are unaffected.
- `blank` in `shouldEnableCustomReasoning` (backend) is treated the same as `none`/`auto` — does not enable custom reasoning.
- No new persistent state format; migration reuses the existing `migrateMap` pattern.

## Testing?

- `npm run lint` — clean.
- `npm run test:unit --prefix tests` — 433 tests pass (27 suites).
- New test cases added to `tests/prompt-converters.test.js` covering `blank` and `xhigh` for `calculateClaudeBudgetTokens` (adaptive + traditional) and `calculateGoogleBudgetTokens` (gemini-3-flash, gemini-3-pro).
- Manual verification not run (no local API keys for xhigh-capable models). The payload path for OpenAI Responses is covered by the existing integration test suite.

## Screenshots (optional)

Not applicable — UI change is a select option rename/addition only.

## Anything Else?

- `xhigh` for xAI is only confirmed on `grok-4.20-multi-agent` per docs. `grok-4.3` tops out at `high`. The frontend filter now allows `reasoning_effort` for both `grok-3-mini` and `grok-4.20-multi-agent`; all other xAI models still strip it.
- The `OPENAI_REASONING_EFFORT_MODELS` allowlist in `src/constants.js` was not changed — it controls which models get the effort field at all on the Azure/direct-OpenAI path, and the existing list already covers the relevant GPT-5.x models.
- Follow-up: UI could display the provider's API default effort for the selected model. Deferred — requires reliable per-model metadata that isn't uniformly available yet.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2178,7 +2178,7 @@
                                                 <a href="https://docs.sillytavern.app/usage/prompts/reasoning/#reasoning-effort" target="_blank" class="opacity50p fa-solid fa-circle-question"></a>
                                             </label>
                                             <select id="openai_reasoning_effort">
-                                                <option data-i18n="openai_reasoning_effort_blank" value="blank">Blank (don't send)</option>
+                                                <option data-i18n="openai_reasoning_effort_none" value="none">None (don't send)</option>
                                                 <option data-i18n="openai_reasoning_effort_minimum" value="min">Minimum</option>
                                                 <option data-i18n="openai_reasoning_effort_low" value="low">Low</option>
                                                 <option data-i18n="openai_reasoning_effort_medium" value="medium">Medium</option>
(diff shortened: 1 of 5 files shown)


platberlitz:
Blank is kinda weird, innit? (british mode) I feel like I'd rather keep Auto.

platberlitz:
Wait, if it's about clarity, why not make it "None" instead of "Auto"? I find blank weird.

TheLonelyDevil9:
@platberlitz fair point, can just change to none, Auto is misleading imo

platberlitz:
I'm good with everything else, changing it to none is just what I wanted.

PR #96, opened 2026-05-14
fix(presets): add connection field guardrails

Geechan:
## Summary
- Extract Chat Completion preset body construction into a pure helper.
- Keep normal OpenAI preset saves from including provider/model/API connection fields.
- Preserve linked preset behavior as an explicit opt-in via `Keep API/model linked to preset`.
- Add regression coverage for excluding connection fields while preserving explicit empty generation values.
## Why
Users reported that linked API/preset behavior can contaminate connection profiles and make provider/model settings unreliable. This keeps normal presets focused on generation/prompt settings, while still allowing advanced linked presets to intentionally save provider/model/API state.
## Notes
- `getChatCompletionPreset()` keeps its legacy default behavior for compatibility.
- `saveOpenAIPreset()` now uses the new helper policy so connection fields are only saved when linked preset mode is enabled.

### DIFF
--- a/public/scripts/openai-preset-utils.js
+++ b/public/scripts/openai-preset-utils.js
@@ -33,3 +33,13 @@ export function getChatCompletionConnectionPresetKeys(settingsMap) {
         .filter(([, [, , , isConnection]]) => isConnection)
         .map(([presetKey]) => presetKey);
 }
+
+/**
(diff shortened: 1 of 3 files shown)


platberlitz:
Tested and worked with my use case. (Changing presets with different connections and switched my profiles around)
Merging

PR #100, opened 2026-05-15
fix(chat): stop chat scroll jumps on macOS browsers

platberlitz:
## Summary
- Adds a generic `is-macos` body class from the parsed browser OS so Safari, Chrome, and Chromium variants on macOS can share platform-specific fixes.
- Disables native CSS scroll anchoring for `#chat` only on macOS browsers.
- Records the chat scroll stability fix in the changelog.

## Root cause
macOS browser-native scroll anchoring can race SillyBunny's JavaScript chat scroll preservation when older messages load or layout settles, causing visible message jumps while scrolling.

## Change
The chat scroller now uses `overflow-anchor: none` under `body.is-macos`, leaving the current non-macOS anchoring behavior unchanged.

## Risk / Rollback
Risk is limited to macOS chat scrolling because the CSS selector is gated by the new body class. Rollback is to remove the `body.is-macos #chat` override and the matching body class helper.

## Verification
- `npm ci --ignore-scripts`
- `npm run lint`
- Not manually verified in macOS Safari/Chrome from this Linux workspace.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -16,6 +16,7 @@ This update consolidates the v1.5.4 staging work since v1.5.3: preset and connec
 
 ### Chat Loading And Search
 - Re-applied chat scroll anchoring across staging and main so scrolling upward no longer skips earlier messages.
+- Disabled native chat scroll anchoring on macOS browsers so it no longer fights SillyBunny's scroll preservation and jumps messages while scrolling.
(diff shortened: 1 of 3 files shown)


TheLonelyDevil9:
Testing

TheLonelyDevil9:
So far, Mac is still a mixed bag in terms of the fix, I'm attempting to fix Chrome, but Safari and Firefox are fine. Merging for now to avoid delays in imminent release

PR #102, opened 2026-05-15
fix(drawer): rectangular character avatar and tag layout

platberlitz:
## Summary
- Makes the character drawer avatar a fixed 2:3 portrait on desktop and mobile, independent of global avatar shape toggles.
- Reworks the drawer controls into a two-column grid so avatar actions and tag input have more breathing room.
- Adds safer tag wrapping spacing and pill height to prevent tag rows from overlapping.

## Verification
- npm run lint


platberlitz:
Does not look right at all

PR #103, opened 2026-05-15
fix: align group controls UI

Geechan:
Fixes alignment issues with group chat controls on desktop. Untested for mobile.

- `rm_group_ai_schedule` box now scales properly to fill empty space.
- `rm_group_automode_delay` and `rm_group_auto_dm_cooldown` boxes now in their own column, preventing text overlap with page size scaling.


platberlitz:
Looks fine on mobile, aligned just fine.

PR #115, opened 2026-05-16
fix: prevent Reasoning Effort overflow on mobile

platberlitz:
## Summary
- Constrain the mobile-only Reasoning Effort dropdown row so it cannot exceed the settings drawer width.
- Force the Reasoning Effort help text to wrap on narrow screens without changing desktop styles.

## Verification
- Ran `npm run check:frontend-budgets` (fails on existing blocking stylesheet budget: 742.5 KiB > 730.0 KiB).
- Confirmed the CSS change is scoped inside the existing `@media screen and (max-width: 760px)` block.

### DIFF
--- a/public/css/sillybunny-mobile-shell.css
+++ b/public/css/sillybunny-mobile-shell.css
@@ -483,6 +483,40 @@
     #right-nav-panel.openDrawer {
         padding-bottom: env(keyboard-inset-height, 0px) !important;
     }
+
+    #left-nav-panel.openDrawer #openai_settings .reasoning-effort-wrapper,
+    #left-nav-panel.openDrawer #openai_settings .reasoning-effort-row {
+        box-sizing: border-box !important;
+        width: 100% !important;
+        min-width: 0 !important;
+        max-width: 100% !important;
+        overflow-x: hidden !important;
+    }
+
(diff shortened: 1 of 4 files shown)


platberlitz:
Follow-up for iOS Safari:
- Added a dedicated `reasoning-effort-row` hook to avoid relying on `:has()`/sibling selectors alone.
- Forced the row, native select, label, and help text to use bounded inline sizes with `!important` where needed to override the base `.oneline-dropdown select { min-width: fit-content; }` behavior on mobile only.

Verification: `npm run check:frontend-budgets` still fails on the existing blocking stylesheet budget (`742.8 KiB > 730.0 KiB`).

platberlitz:
The repeated iOS Safari repro is very likely cached CSS: `sillybunny-theme.css` changed, but `index.html` was still requesting `css/sillybunny-theme.css?v=20260516b`, so Safari could keep serving the old stylesheet. I pushed a cache-key bump to `v=20260516c` so the mobile width fix is actually fetched after update.

Verification: `npm run check:frontend-budgets` still fails on the existing blocking stylesheet budget (`742.8 KiB > 730.0 KiB`).

platberlitz:
Fuck's sake it won't work

platberlitz:
I'm gonna redo because this is NOT working

PR #120, opened 2026-05-16
fix: refresh topbar tokens after hiding messages

platberlitz:
## Summary
- emit MESSAGE_UPDATED after hide/unhide operations complete and save
- refresh topbar context tokens immediately when MESSAGE_UPDATED fires

## Validation
- npx eslint public/scripts/chats.js public/scripts/sillybunny-tabs.js
- git diff --check


platberlitz:
Tested and working.

PR #121, opened 2026-05-17
Fix Docker startup regressions

platberlitz:
## Summary
- Prevent bundled assistant cards from being recreated on startup after the user deletes them
- Namespace SillyBunny session cookies to avoid collisions with SillyTavern on the same host
- Add default PUID/PGID compose environment values for host-friendly file ownership
- Add common Docker/private network CIDRs to the tracked default whitelist template

## Verification
- bunx eslint public/scripts/welcome-screen.js src/users.js
- node --check public/scripts/welcome-screen.js
- node --check src/users.js
- bun YAML parse check for default/config.yaml and docker/docker-compose.yml
- git diff --check

Note: docker is not installed in this environment, so I could not run docker compose config.

### DIFF
--- a/default/config.yaml
+++ b/default/config.yaml
@@ -64,6 +64,8 @@ enableForwardedWhitelist: true
 whitelist:
   - ::1
   - 127.0.0.1
+  - 172.16.0.0/12
+  - 192.168.0.0/16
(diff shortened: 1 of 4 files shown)


platberlitz:
Didn't get errors

PR #123, opened 2026-05-17
fix(quick-replies): dedupe and force-update bundled Memory Sharding preset

platberlitz:
## Summary
- Add a managed bundled preset sync for the Memory Sharding Quick Reply set during startup content seeding.
- Replace stale or duplicate installed copies with the current bundled preset, keeping a single canonical file.
- Respect user deletion by doing nothing when no installed copy exists.

## Verification
- Ran targeted scratch startup checks for duplicate cleanup, stale single-copy replacement, idempotency, and deletion-respect behavior.
- Ran targeted ESLint on src/endpoints/content-manager.js.
- Confirmed package.json has no npm test script.

### DIFF
--- a/public/scripts/extensions/quick-reply/index.js
+++ b/public/scripts/extensions/quick-reply/index.js
@@ -33,6 +33,8 @@ const defaultSettings = {
 
 /** @type {Boolean}*/
 let isReady = false;
+/** @type {Boolean}*/
+let initStarted = false;
 /** @type {Function[]}*/
 let executeQueue = [];
 /** @type {string}*/
@@ -168,6 +170,12 @@ const handleCharChange = () => {
 };
 
 export async function init() {
+    if (initStarted) {
+        debug('init() already started; skipping duplicate call');
+        return;
+    }
+    initStarted = true;
+
     await loadSets();
     await loadSettings();
     log('settings: ', settings);
--- a/public/scripts/extensions/quick-reply/index.js
+++ b/public/scripts/extensions/quick-reply/index.js
@@ -33,8 +33,8 @@ const defaultSettings = {
 
 /** @type {Boolean}*/
 let isReady = false;
-/** @type {Boolean}*/
-let initStarted = false;
+// SillyBunny divergence: shared in-flight init promise; resets on rejection.
+let initPromise = null;
 /** @type {Function[]}*/
 let executeQueue = [];
 /** @type {string}*/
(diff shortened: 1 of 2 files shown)


TheLonelyDevil9:
((All GPT-5.5 btw)):

Thanks for the work here. I'm requesting changes before merge because the new managed bundled preset sync in `src/endpoints/content-manager.js` has a user-data-loss risk.

It appears to match Quick Reply files by JSON `name` and delete/replace all matches. Since the managed list includes both `Default` and `Memory Sharding` while the PR scope is `Memory Sharding`, this can silently clobber user-customized `Default` or `Memory Sharding` quick-reply presets on startup.

Please narrow the sync to known stale bundled copies, a known filename+hash, a one-shot migration marker, or another mechanism that preserves/renames user-modified copies instead of overwriting user-authored presets. The visible duplicate `Memory Sharding` issue still needs handling, and the bundled prompt should be updated per the follow-up.

Minor: the duplicate init guard should preferably share an in-flight promise and reset on rejection, and the base file should include a SillyBunny divergence comment.

CI is green, but targeted regression tests or scratch checks should cover stale copy replacement, dedupe, deletion-respect, and preserving customized copies. Request changes before merge.

platberlitz:
> ((All GPT-5.5 btw)):
> 
> Thanks for the work here. I'm requesting changes before merge because the new managed bundled preset sync in `src/endpoints/content-manager.js` has a user-data-loss risk.
> 
> It appears to match Quick Reply files by JSON `name` and delete/replace all matches. Since the managed list includes both `Default` and `Memory Sharding` while the PR scope is `Memory Sharding`, this can silently clobber user-customized `Default` or `Memory Sharding` quick-reply presets on startup.
> 
> Please narrow the sync to known stale bundled copies, a known filename+hash, a one-shot migration marker, or another mechanism that preserves/renames user-modified copies instead of overwriting user-authored presets. The visible duplicate `Memory Sharding` issue still needs handling, and the bundled prompt should be updated per the follow-up.
> 
> Minor: the duplicate init guard should preferably share an in-flight promise and reset on rejection, and the base file should include a SillyBunny divergence comment.
> 
> CI is green, but targeted regression tests or scratch checks should cover stale copy replacement, dedupe, deletion-respect, and preserving customized copies. Request changes before merge.

Implemented and pushed a follow-up fix in 7a9be45.

This replaces the name-based bundled quick-reply sync with hash-gated reconciliation scoped only to Memory Sharding. User-authored or edited files are now ignored unless their bytes match a known bundled hash, deletion is respected, and staleHashes is left empty for this PR so the change only dedupes existing identical bundled copies. I also replaced the quick-reply initStarted boolean with a shared in-flight promise that resets on rejection, so duplicate init calls share the same work while failed init can still retry.

Verified with the planned scratch reconciliation scenarios and ESLint.

PR #125, opened 2026-05-17
fix(agents): prevent duplicate Prose Polisher after template prompt update

platberlitz:
## Summary
- Prompt before adding another copy of a template-backed agent when the template id is already installed.
- Heal existing same-template duplicate agents on extension load, preferring the current bundled template prompt and preserving phase-locked agents.
- Show an Added pill and dimmed state for already-installed templates in the Templates browser.

## Verification
- npm run test:unit -- in-chat-agents-store.test.js in-chat-agents-templates.test.js
- npx eslint public/scripts/extensions/in-chat-agents/agent-store.js public/scripts/extensions/in-chat-agents/index.js
- ./node_modules/.bin/eslint in-chat-agents-store.test.js (from tests/)
- git diff --check


platberlitz:
Tested and works on my end.

PR #126, opened 2026-05-17
feat: categorise templates browser, retire Pathfinder agent

platberlitz:
Problem

The In-Chat Agents Templates modal showed built-in templates as one flat grid, which made tracker-heavy catalogs hard to scan and left useful metadata such as tags, categories, versions, and append execution order mostly hidden. Pathfinder was also still exposed as an in-chat agent template despite having its own Extensions tab surface.

Approach

This PR adds modal-only template subcategories, groups the browser by category/subcategory with search and single-select filter pills, adds template version badges, and retires Pathfinder from the in-chat agent template/default catalog while keeping its runtime code intact. Agent cards now surface execution order and template versions, including an update-available affordance that refreshes bundled agents from their source template while preserving key user state.

Screenshots

Not attached from this terminal pass. I smoke-tested the dev server at http://127.0.0.1:4444/ after implementation; the intended manual screenshot set is default browser view, filtered view, 720px responsive view, and version-update affordance.

Tests run

- npm ci --ignore-scripts
- npm ci --ignore-scripts --prefix tests
- npm run lint
- npm run lint --prefix tests
- npm run test:unit --prefix tests -- tests/in-chat-agents-templates.test.js
- git diff --check

### DIFF
--- a/public/scripts/extensions/in-chat-agents/agent-store.js
+++ b/public/scripts/extensions/in-chat-agents/agent-store.js
@@ -649,6 +649,19 @@ export const AGENT_CATEGORIES = {
     custom: { label: 'Custom', icon: 'fa-puzzle-piece' },
 };
 
+/**
+ * Modal-only template subgroup labels.
+ */
+export const AGENT_SUBCATEGORIES = {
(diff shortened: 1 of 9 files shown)


platberlitz:
Tested + working

PR #129, opened 2026-05-17
fix: pre-gen intercept running toast and structured edits view

platberlitz:
## Summary
- Show a cancelable running toast while pre-generation intercept agents are rewriting context
- Render chat-format intercept history as structured per-message changes instead of raw JSON walls
- Keep text-format intercept history on the existing prompt diff path and collapse raw output in details
- Preserve invalid chat intercept output for fallback history diffs

## Tests
- npm run lint
- npm run lint --prefix tests


platberlitz:
Tested + working.

PR #131, opened 2026-05-17
fix: strip reasoning blocks from profile-based prompt transform output

platberlitz:
Text-completion reasoning models (DeepSeek-R1 family, Qwen3-thinking, etc.) emit `<think>...</think>` inline inside `choices[0].text`. The profile-based path used by post-gen agents, `refineLLMCall`, and the Pathfinder sidecar pulled that raw text via `extractProfileResponseText` and surfaced it verbatim, so the model's reasoning leaked into the chat-visible message. The non-profile path already strips reasoning via `generateQuietPrompt({ removeReasoning: true })`.

Fix: route the helper's return through the existing `removeReasoningFromString` utility so all profile-path callers get parity with the main-model path. Honors auto-parse settings (returns input unchanged if the user has disabled reasoning auto-parse).

Test maintenance: update the affected Jest mocks so the runner test stays Node-friendly after `llm-utils.js` imports the browser-side reasoning module, and so the pre-intercept summary mock matches the current `agent-store.js` exports.

Verification:
- `npm run test:unit -- --runInBand tests/in-chat-agents-pre-intercept-summary.test.js tests/in-chat-agents-runner.test.js`
- `npm test -- in-chat-agents` runs all 33 Jest suites successfully (492 tests), then exits nonzero because Playwright reports `No tests found` for the `in-chat-agents` selector.

### DIFF
--- a/public/scripts/extensions/in-chat-agents/llm-utils.js
+++ b/public/scripts/extensions/in-chat-agents/llm-utils.js
@@ -1,7 +1,8 @@
 import { normalizeContentText } from '../../../script.js';
+import { removeReasoningFromString } from '../../reasoning.js';
 
 export function extractProfileResponseText(response) {
-    return normalizeContentText(response?.content)
+    const text = normalizeContentText(response?.content)
         || normalizeContentText(response?.choices?.[0]?.message?.content)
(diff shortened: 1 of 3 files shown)


platberlitz:
Tested and works

PR #134, opened 2026-05-17
fix: tighten advanced formatting mobile header

platberlitz:
## Summary
- Hide Advanced Formatting master import/export labels on mobile so the header does not overlap
- Slightly reduce the mobile header gap for the Text Completions Advanced Formatting drawer

## Testing
- git diff --check


platberlitz:
Checked and confirmed it doesn't affect desktop

Geechan (approved):

PR #143, opened 2026-05-19
feat(cards): add opt-in card script sandbox runtime

platberlitz:
## Summary
- add the card script runtime bridge for sandboxed iframe execution
- wire the opt-in message action and enabled setting into the main UI
- validate postMessage source, nonce, messageId, schema, allowlist, and rate limits before executing slash commands
- add unit and Playwright E2E coverage plus changelog/security invariant notes

## Security
- execution stays off by default and requires a per-message user action
- iframe sandbox uses `allow-scripts` only, without `allow-same-origin`
- sandbox receives no command results, chat contents, settings, or parent DOM access
- default runtime policy stays on the narrow allowlist tier

## Verification
- `npm run lint`
- `npm run lint --prefix tests`
- `npm run test:unit --prefix tests`
- `npm run test:e2e --prefix tests -- card-script-sandbox.e2e.js`
- `git diff --check`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@ This update consolidates the v1.6.0 staging work since v1.5.3: preset and connec
 
 ### Character Cards
 - fix(cards): warn when card HTML contains stripped `<script>`/`<iframe>` blocks (#94).
+- feat(cards): add opt-in sandboxed execution for supported card scripts (#94).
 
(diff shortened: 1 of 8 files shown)


platberlitz:
Working on the safest native JS sandbox after this.

PR #145, opened 2026-05-19
fix: constrain API settings panel width on laptop screens

platberlitz:
## Summary
- Adds laptop-width constraints for the Workspace/API and Customize shells from 1001px to 1440px wide
- Updates runtime shell sizing so inline `!important` dimensions follow the new constrained width instead of forcing full width on short laptop screens
- Bumps `sillybunny-tabs` JS/CSS asset query strings so browsers load the fix

## Test Plan
- [x] `npx eslint public/scripts/sillybunny-tabs.js`
- [x] CSS parsed with `@adobe/css-tools`
- [x] `git diff --check`
- [x] Browser-measured shell widths at 1001x768, 1366x768, and 1440x900 with saved shell size temporarily cleared and restored

### DIFF
--- a/public/css/sillybunny-tabs.css
+++ b/public/css/sillybunny-tabs.css
@@ -6160,6 +6160,15 @@ body #right-nav-panel #HotSwapWrapper .hotswap > :is(.avatar, .character_select,
     }
 }
 
+@media screen and (min-width: 1001px) and (max-width: 1440px) {
+    #left-nav-panel.openDrawer,
+    #user-settings-block.openDrawer {
+        width: min(100vw, clamp(680px, 60vw, 920px)) !important;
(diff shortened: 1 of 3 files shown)


platberlitz:
Expanded this PR with the compact-mode follow-up:

- Added compact-mode laptop viewport sizing for the Workspace/API settings panels at 1001px-1440px, scaling the PR #145 constraints down to `clamp(600px, 58vw, 810px)` with a 24px viewport gutter.
- Removed the redundant Workspace subtitle so the panel header stays cleaner.
- Kept the CSS as a flat media query for compatibility with the repo's existing browser-support style.

Verified with `npx eslint public/scripts/sillybunny-tabs.js` and `npm run build:frontend`.

PR #152, opened 2026-05-21
fix: stop generation before cleanup listeners lose state

TheLonelyDevil9:
## What?
Fixes the stop-generation path so clicking Stop emits the stop lifecycle before cleanup clears the active streaming processor. Also updates Guided Generations swipe waiting so it resolves on either normal generation end or generation stop.

## Why?
Stop-generation listeners, especially in-chat agent cleanup, need access to the active streaming state while handling `GENERATION_STOPPED`. The old order cleared and unblocked first, which could make stop listeners miss the state they needed and allow generation-related work to continue after Stop was clicked.

## How?
`stopGeneration()` now emits `GENERATION_STOPPED` synchronously before clearing the current streaming processor, then unblocks the UI without also faking a normal `GENERATION_ENDED` event. The send-button cleanup helpers keep their previous default behavior for normal completion but can suppress `GENERATION_ENDED` for the stop path. Guided swipe now waits for either stopped or ended so the changed stop semantics do not leave it pending.

## Testing?
- `npm run test:unit --prefix tests -- event-emitter.test.js in-chat-agents-runner.test.js`
- `npx eslint public/script.js public/scripts/extensions/guided-generations/scripts/guidedSwipe.js tests/event-emitter.test.js tests/in-chat-agents-runner.test.js`

## Screenshots (optional)
Not included; this is lifecycle/event-order behavior rather than a visual UI change.

## Anything Else?
Requesting Platberlitz review before merge. During pre-review, Guided Generations swipe was identified as a stop/ended event consumer and patched so it does not hang when a generation is stopped.

### DIFF
--- a/public/script.js
+++ b/public/script.js
@@ -4426,12 +4426,14 @@ function showStopButton() {
     $('#mes_stop').css({ 'display': 'flex' });
 }
 
-function hideStopButton() {
+function hideStopButton({ emitGenerationEnded = true } = {}) {
     $('#send_form').removeClass('sb-generating-controls');
     // prevent NOOP, because hideStopButton() gets called multiple times
     if ($('#mes_stop').css('display') !== 'none') {
         $('#mes_stop').css({ 'display': 'none' });
-        eventSource.emit(event_types.GENERATION_ENDED, chat.length);
+        if (emitGenerationEnded) {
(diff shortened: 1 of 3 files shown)


platberlitz (approved):
Looks good and a simple fix

PR #153, opened 2026-05-21
feat: refactor character editor drawer

Geechan:
Completely reworks the character editor menu to improve QOL, layout, and ease of use.

- Character menu no longer closes when opening a new character. Instead, the reply bar and selected character flash green to let you know they've been selected, alongside highlighting the editor icon. On mobile, the old behaviour persists, and the character menu will auto-close when selecting a new character.
- When generating a model response, the reply bar will glow yellow.
- Character editor has been completely overhauled. All character definitions have been split into four sub-tabs for navigation - Information, Definitions, Greetings, and Metadata.
- ADV. definitions window has been depreciated in favour of including all definitions in the new Definitions tab.
- Added a full screen option for the character editor, which removes everything above the new sub-tab navigation.
- World Info lorebooks has moved from Workspace to Characters menu. This aligns it better with its actual use case.

Note: has not been tested on mobile, but will likely work.


platberlitz:
Pushed a small mobile-only CSS follow-up for the character editor drawer:\n\n- Metadata tab: keeps Created by / Character Version compact and gives Creator's Notes / Tags more vertical room.\n- Definitions tab: aligns the Character's Note row with the @ Depth / Role / Tokens column on mobile.\n- Greetings tab: makes the alternate greeting delete control icon-only so the header does not overflow.\n\nVerified the stylesheet parses cleanly and the local server serves the appended mobile block.

platberlitz (approved):

platberlitz:
Double-checked and does not mess with desktop. Mobile looks fine.

PR #157, opened 2026-05-22
chore: remove redundant fork update launcher

platberlitz:
## What?

- Remove the legacy Windows update launchers `UpdateAndStart.bat` and `UpdateForkAndStart.bat`.
- Remove their remaining references from `Update-Instructions.txt` and `.github/pr-auto-labels-by-files.yml`.
- Leave `Start.bat` as the documented Windows launcher path.

## Why?

The standalone update launchers duplicated behavior that now belongs to the normal launcher/self-update path. Keeping the extra scripts around makes the Windows update surface harder to maintain and easier for users or developers to pick the wrong path.

## How?

- Deletes both legacy update launcher scripts instead of moving fork/update logic into another standalone launcher.
- Keeps the existing `Start.bat` flow as the Windows entrypoint.
- Cleans up docs and PR auto-label file matching so they no longer reference deleted scripts.

## Testing?

- `git diff --check HEAD~1..HEAD`
- `git grep -n "UpdateForkAndStart" HEAD -- ':!releases/*.zip'` returned no matches before the follow-up removed `UpdateAndStart.bat` too.
- CI is green: merge conflict check, tracked default-user state check, ESLint, unit tests, frontend budgets, semantic PR, and merge-blocking-label check.

## Screenshots (optional)

Not applicable; this is launcher/docs cleanup with no UI surface.

## Anything Else?

Reviewer note: this is intentionally a user-facing launcher removal. Merge only if `Start.bat` is now the accepted Windows update/start path for Git installs.


platberlitz:
Follow-up: removed `UpdateAndStart.bat` too. I also removed its remaining live references from `Update-Instructions.txt` and `.github/pr-auto-labels-by-files.yml`, leaving `Start.bat` as the Windows launcher path documented here.

PR #277, opened 2026-05-29
fix: defer iOS streaming scroll during touch

platberlitz:
## Summary
- defer streaming bottom pins while mobile touch, iOS momentum, or user-scroll lock is active
- stop mobile streaming autoscroll after real user movement until the user returns to bottom
- defer live streaming DOM writes during mobile gestures so fast iOS scrolling is not pulled by reflow
- preserve viewport only when it will not fight an active gesture

## Tests
- npm run test:unit -- --runTestsByPath chat-render-lifecycle-scroll-intent.test.js chat-render-lifecycle-bottom-scroll.test.js chat-render-lifecycle-script-wiring.test.js mobile-streaming.test.js
- npx eslint public/script.js public/scripts/chat-render-lifecycle/scroll-intent.js
- npx eslint chat-render-lifecycle-scroll-intent.test.js chat-render-lifecycle-script-wiring.test.js
- git diff --check -- public/script.js tests/chat-render-lifecycle-script-wiring.test.js

## Manual testing
- Needs latest physical iOS verification on this branch.

### DIFF
--- a/public/script.js
+++ b/public/script.js
@@ -643,6 +643,7 @@ let mobileStreamingBottomPinTimer = 0;
 let mobileStreamingBottomPinSettle = false;
 let mobileStreamingBottomPinSmooth = false;
 let deferredMobileStreamingBottomPin = false;
+let deferredMobileStreamingBottomPinTimer = 0;
 let lastMobileStreamingBottomPinAt = 0;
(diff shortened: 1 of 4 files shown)


platberlitz:
Didn't work well

PR #282, opened 2026-05-31
fix: persist sampler visibility choices

platberlitz:
## Summary
- auto-enable and immediately save manual sampler priority when a Text Completion sampler is toggled
- persist the Prioritize toggle immediately instead of waiting for popup close
- hide repetition/frequency/presence penalties, Mirostat, and XTC controls by default unless manual sampler selection is active

## Testing
- npm run lint -- public/scripts/samplerSelect.js public/scripts/textgen-settings.js
- git diff --check
- npm run build:frontend


platberlitz:
bwee

PR #331, opened 2026-06-03
fix: align docker compose build context

TheLonelyDevil9:
## What?
Update the Docker Compose build configuration so `docker/docker-compose.yml` resolves the SillyBunny build context from the repository root. The Docker launch command in both README copies now includes `--project-directory .` to match that path model.

## Why?
Coolify deploys the compose file with `--project-directory` set to the repository root. The previous `build: ..` resolved one directory above the checkout in that invocation, so Docker could not find the root `Dockerfile` and the deployment failed before building.

## How?
The compose service now uses an explicit build object with `context: .` and `dockerfile: Dockerfile`. Bind mounts now point at `./docker/config`, `./docker/data`, `./docker/plugins`, and `./docker/extensions` so runtime Docker data still lives under the Docker folder when the project directory is the repository root.

## Testing?
- `git diff --check`
- Verified the Coolify-style path model resolves to the repository root and root `Dockerfile`.
- Docker build was not run because Docker CLI is unavailable in this environment.

## Screenshots (optional)
Not applicable; deployment configuration change only.

## Anything Else?
This is targeted at `main`, where the Coolify deployment failure occurred.

### DIFF
--- a/.github/readme.md
+++ b/.github/readme.md
@@ -95,7 +95,7 @@ Then, run the appropriate launcher for your OS, which auto-installs all dependen
 | macOS (Terminal) | `./Start.command` |
 | macOS (Finder) | Double-click `Start.command` (right-click > Open if Gatekeeper warns) |
 | Linux / WSL | `./start.sh` |
-| Docker | `docker compose -f docker/docker-compose.yml up --build`
+| Docker | `docker compose --project-directory . -f docker/docker-compose.yml up --build`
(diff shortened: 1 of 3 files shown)


TheLonelyDevil9:
Closed in favor of #332, which targets staging per repository PR policy and carries the same Docker Compose path fix without the main-vs-staging diff noise.

PR #446, opened 2026-06-12
feat: companion agents — sidecar-style auxiliary AI cards

platberlitz:
## Companion Agents

Companion Agents are extra AI helpers that run alongside your main chat. While you talk with your character, a companion quietly reads the conversation and does its own job in the background: tracking the plot, watching how relationships develop, drafting worldbuilding notes, and so on.

Each companion lives as a note card in your chat, or quietly in the side panel if you prefer. You can check what it's doing at any time, or just let it work on its own.

### What can they do?

Some examples:

- Noting plot holes, contradictions, or unresolved story threads
- Tracking how character relationships and trust levels shift over time
- Drafting new lorebook entries from facts the scene establishes
- Compressing old chat history so earlier context stays accessible
- Adding narrator-style commentary on what's happening in the scene
- Generating live audience reactions to the latest reply
- Rendering in-story messages (texts, letters, scrolls) as a styled inbox card
- Holding a private side chat with the characters, kept out of the story

### How do they run?

Each companion has one of two trigger modes:

- Auto: runs automatically after matching AI replies
- Manual: you trigger it yourself whenever you want

You can also add keyword or regex conditions and a probability percentage if you only want a companion to fire some of the time or on specific patterns.

### What does it look like?

Companions show up as note cards in your chat. While a companion is running, the card shows a spinner. Once done it shows the note, or an error message if something went wrong. Each card has four buttons: regenerate, edit the note by hand, copy, and delete.

Some companions are interactive right from the card. Chat Only has a text box for typing an aside, and Plot Compass has a box for setting where you want the story to go. Both also work from the side panel.

Some companions are configured to only appear in the side panel rather than as inline cards.

### The panel

A draggable handle lets you open the companion panel at any time. You can drop the handle onto any screen edge (left, right, top, or bottom) and it stays docked there. From the panel you can regenerate companions, edit notes by hand, jump to the source message, and view recent run history per companion. A lock button keeps the panel open when you click outside it. A regenerate-all button re-runs every companion on the last reply at once.

Enabled companions show up in the panel right away, even before their first run, so you can see what is on and open settings without waiting.

### The dashboard

There is a dedicated Companion Agents page on your Agents dashboard. From there you can:

- See every companion across all your chats in one place
- Create new ones, import from a file, or pick from bundled templates
- Enable or disable companions with a single tap
- See which of your existing inline agents can be switched to companion execution

### Highly customizable

Just about everything on a companion can be tuned per companion:

- Trigger: auto or manual, with optional keyword/regex conditions and a probability percentage
- Display: inline note card, side panel only, or hidden (feeds future runs without showing)
- Output format: markdown, HTML, or plain text
- Context window: how many recent messages to include (1 to 50), how deep to read history (1 to 10), and a minimum token threshold before auto companions kick in (for example, Memory Shard waits until around 30k)
- Max tokens per run
- What to feed the companion: character card, persona, world info, author's note, system prompt, and prior companion notes, each toggled on or off
- Feedback: let a companion read its own recent notes to stay consistent, with adjustable depth
- Run with enabled companions in one request: when companions share the same profile, model, and context settings, they can share a single model request
- Raw prompt mode: use the agent's prompt exactly as written, with no extra scaffolding
- Per-companion model, temperature, and reasoning settings

Several companions also have their own extras, like the Director's selectable narration voices, the Plot Compass objective, and the Chatroom's selectable chat styles.

### A note on models

It is recommended to give companions their own smaller model. Background jobs like summarizing or tracking relationships do not need your main chat model, and smaller models handle them just fine while costing less and running faster. Your main chat can stay on whatever model you prefer.

Companions can also run in parallel (faster) or sequentially (easier on rate limits), set in the extension settings.

### What's included

This ships 11 ready-to-use companions you can start with:

- Continuity Companion: notes contradictions, unresolved setup, and state changes that may matter in the next scene
- Relationship Lens: notes subtext, trust shifts, leverage, and emotional consequences between characters
- Director's Commentary: adds narrator-style commentary on the scene using a selectable built-in narration voice (or a custom one you write yourself)
- Actor Interview: interviews your character about the latest scene, in character and off the record
- Lorebook Scout: reads the scene for new durable world facts (places, factions, NPCs, customs) and drafts lorebook entries with trigger keys; skips facts already covered by your World Info
- Memory Shard: compresses older chat history into a compact shard once the chat reaches around 30k tokens, feeds it back into context, and can optionally hide the messages it summarized
- NPC Motivator: helps NPCs feel less static by surfacing their goals and giving them more agency
- Plot Compass: you set a Plot Objective right on the card or in settings, and the companion tracks progress toward it, flagging next beats, obstacles, and a suggested move each turn
- Chatroom: generates live audience reaction notes for the latest reply, styled as a chat feed with selectable chat styles
- Message Inbox: watches for in-story messages sent to your character (texts, DMs, letters, scrolls) and renders them as a styled card; stays hidden when the reply has none
- Chat Only: a private side chat with the characters in the scene. Type an aside into the box and send, and the present characters reply in voice, candidly and out of the main story. It keeps a running side-chat transcript across turns

### Other notes

- Auto cleanup: companions tied to a specific chat are removed automatically when you delete that chat.


platberlitz:
Works strangely. Redoing

platberlitz:
Nevermind, there's something here...

platberlitz:
## What changed in this push (7 commits, `b39dc56dd..1a1030312`)

This round addresses three reported gaps — converting an agent to Companion **deleted the entire message history**, there was no one-click conversion, and no menu opened a dedicated Companion UI — plus the remaining parity items from the original feature plan.

---

### 1. `fix(ica): stop extra-only snapshot refreshes from reloading the chat` — the history-wipe bug

**Mechanism:** saving an agent calls `refreshRegexSnapshotsForAgent()`, which walks every chat message. A converted (now-companion) agent is excluded from `isRegexRefreshAgentCandidate`, so the regex snapshot stored in each message's `extra` gets deleted — on **every** message. Each deletion scheduled a refresh; for any message outside the lazy-render window (i.e. the whole older history of a long chat) `refreshMessageAfterMutation` fell back to `saveChat()` + `reloadCurrentChat()`. `clearChat({clearData:true})` empties the in-memory `chat` mid-reload, and an interrupted/raced reload leaves an empty chat that any later save persists. Disabling or deleting an inline regex agent triggers the same cascade on staging today; conversion just made it deterministic.

**Fix (`agent-runner.js`):** snapshot updates are bookkeeping-only `extra` mutations — they can't change how an unrendered message will render later. The two snapshot-only call sites (`ensureMessageRegexSnapshot`, `refreshRegexSnapshotForAgentOnMessage`) now pass `skipReloadFallback: true` through `scheduleMessageRefresh`: rendered messages still update in place via `updateMessageBlock`, off-screen messages do nothing (persistence is already arranged by the callers). The per-index refresh dedupe merges flags so a pending *genuine text mutation* refresh is never downgraded by a later bookkeeping one. All other `scheduleMessageRefresh` call sites (transform undo/redo, tracker fixes, manual runs) keep the existing save+reload fallback — covered by a guard test.

**Reviewer note:** the `saveChat` expectation in the existing test *“refreshes existing regex snapshots when an agent regex changes”* changed accordingly (no more full save+reload for an off-screen snapshot rewrite). 5 new regression tests, including a 3-message conversion that asserts identical message objects, intact `mes`, snapshots stripped from both `extra` and `swipe_info`, and zero `saveChat`/`reloadCurrentChat` calls.

### 2. `feat(ica): one-click inline/companion conversion on agent cards`

- New store helper `convertAgentExecution(agent, target)` (`agent-store.js`): → companion keeps prompt/regex/injection/conditions, normalizes the companion config, sets `phase: 'post'`; → inline moves `category: 'companion'` agents to `custom` because `normalizeAgent` re-derives execution from category (otherwise the agent would snap back to companion on save). `agent.companion` is kept so conversions round-trip. Tool agents refuse.
- Every agent card gets a convert action (`.ica--btn-convert-execution`): fa-user-astronaut on inline agents, fa-right-left on companions; the toast mentions the category move when it happens. Select mode gains a bulk **To Companion** button (`#ica--bulkConvertCompanion`).
- The conversion flow funnels through one `applyAgentExecutionConversion()` (lock template customization → save → snapshot refresh → tool re-sync → re-render), which is also what the dashboard hook uses.

### 3. `feat(ica): companion dashboard with wand menu and toolbar entry`

New `companion/companion-dashboard.js` (~350 lines), opened from a **wand-menu item** (`#ica_companions_wand_item`, appended to `#extensionsMenu` at runtime — no upstream template change, so no extra ledger entry) and a **Companions** toolbar button (`#ica--companionsDashboard`) next to Fix Trackers. Large popup with:

- companion rows: enable toggle, trigger/display/format/batch/feedback pills, **Run on last reply**, **Edit**, **Convert to inline**;
- a *Convert an existing agent* section listing eligible inline agents with one-click conversion;
- **Run All on Last Reply**, **New Companion** (editor pre-set to companion), and **AI Maker** (editor + auto-opened maker);
- a *Recent notes* browser (last 20 done notes, newest first) with jump-to-message;
- a global-disabled notice when In-Chat Agents are off.

It re-renders live on `COMPANION_RESULTS_UPDATED_EVENT` while open (debounced, scroll preserved, listener removed on close) and receives all index.js-owned behavior through a `configureCompanionDashboard(hooks)` seam — no import cycle. `openEditor()` gained `{ draft, autoOpenCompanionMaker }` for the create flows. Mobile rules live in the existing 768px media block; all touch targets ≥32px.

### 4. `fix(ica): apply agent regex to companion cards and missing render events`

- `formatCompanionContent` now runs the owning agent's regex scripts (`applyRegexScriptList`, `AI_OUTPUT` placement, same options as the chat display path in `script.js`) **before** the markdown/html/text format step, so a tracker converted to companion keeps its beautifier inside the card. Sanitization (DOMPurify / escaping) still happens after the regex pass; deleted agents degrade to raw content.
- Card rendering now also re-runs on `MORE_MESSAGES_LOADED` (cards were silently missing from lazy-loaded history), `MESSAGE_EDITED`, and `MESSAGE_DELETED` — the delete handler re-renders all messages because indices shift, instead of trusting the deleted index payload.

### 5–6. Four new bundled companion templates (catalog now has six)

| Template | Trigger | Notable config |
|---|---|---|
| Director's Commentary | auto | batched scene commentary, 8-msg context |
| Actor Interview | manual | in-character backstage interview, character card included |
| Lorebook Scout | manual | drafts **one-concept** lorebook entries with `Keys:` lines; World Info included so it skips already-covered lore |
| Memory Shard | manual | token-lean distillation of the bundled *Memory Sharding* QR — keeps its minimum reconstruction set (TIMELINE / CHARACTERS / RELATIONSHIPS / weighted EVENTS / DIALOGUE KEYS / THREADS / NOW) in the same terse pipe formats, drops the 17-section protocol and examples; 30-msg context, builds on prior shards via note history, and **feedback-injects the latest shard** into the main prompt as rolling memory |

(An art-prompt template was considered and dropped — the Image Gen extension already covers that ground.) All ship disabled, byte-identical between source files and `templates/index.json` (enforced by the templates test).

### 7. `test(ica)` + `docs`

- New coverage locking in behavior that existed but was untested: companion keyword triggers scan the **latest user message** (incl. `continue`, where the chat tail is the assistant reply), assistant-only mentions don't activate, regex-literal triggers (`/dragon\s+lair/i`) compile, and notes mirror into `swipe_info` so they survive swipe-away/swipe-back and delete from both layers.
- `docs/upstream-touch-ledger.md`: new entry for the `public/index.html` static `mes_run_companions` button row (hidden div only; all logic in the extension).

---

### Test state

`cd tests && npm run test:unit -- in-chat-agents`: **7 suites / 115 tests green** (was 5 / 92 at the original draft head — +23 tests, 2 new suites: `in-chat-agents-companion.test.js`, `in-chat-agents-companion-dashboard.test.js`). Root-level suites that need server deps fail in a fresh worktree (`Cannot find module 'express'`) — environmental; CI covers them.

**Still to smoke manually before undrafting:** long-chat (100+ msgs) conversion in a real browser (no reload flicker, history intact after F5), dashboard interactions on a 375px viewport, and a feedback-enabled companion showing its note in the next prompt preview.

platberlitz:
## Follow-up: slide-out Tracker Panel (`19e1c…` / latest commit)

User feedback on the converted-tracker experience: tracker state was rendering as a card **inside the chat flow**, and the standalone companion call was writing scene prose around the state block instead of just tracking. Both addressed:

### `panel` display mode + slide-out Tracker Panel

- Companions gain a third display mode — `card` / **`panel`** / `hidden` (editor select updated, `normalizeCompanionConfig` allows it). `panel` results are stored and still usable for feedback injection, but **never render as cards under messages**.
- **Converting a tracker to companion now defaults to `panel`** — its state leaves the chat flow entirely. The conversion toast says where the state went; the editor can switch back to cards.
- New `companion/companion-panel.js`: a fixed right-edge slide-out panel showing each companion's **latest rendered state** (same regex + format pipeline as cards), with regenerate, jump-to-source-message, and a collapsed *Previous states* history (last 5). Live-updates on companion results and chat changes.
- Three ways in: a slim right-edge **Trackers handle** (auto-hidden when there's nothing to show), a **Tracker Panel** wand-menu item, and a button in the Companions dashboard. All injected at runtime — no new upstream-file touches.

### Tracker companions stop writing scenes

Tracker prompts were written to ride along with a story reply, so running them standalone made models add narration around the state block. `buildCompanionPromptMessages` now appends a system instruction for tracker-category companions: *state only, exact format, no narration/dialogue/prose*.

### Header polish

- Companion results now store the connection profile's **display name**; unresolvable profiles store an empty label instead of the raw id, and previously-stored raw uuids are suppressed at render time.
- Agent names carrying a trailing uuid (imported/AI-generated agents) are cleaned in card and panel headers (`cleanCompanionAgentName`) — stored data untouched.

### Tests

8 suites / **127 tests** green (new `in-chat-agents-companion-panel.test.js`; runner suite covers the state-only instruction and profile-label resolution; store suite covers panel normalization and the tracker→panel conversion default; UI suite covers panel/hidden ledger exclusion and name cleanup).

platberlitz:
## Follow-up 2: the automatic tracker loop (`bc25e4887`)

Direct user feedback drove this round: tracker cards were still rendering in the chat (pre-existing agents never got the new defaults), the card chrome degraded to unstyled defaults on some themes/browsers, the appended format instructions were distorting tracker prompts, and the panel needed to be reachable from the chat screen without detours.

### The loop is now the default for trackers

Converting a tracker to companion execution (and a **one-time migration** for already-converted tracker companions, gated by a `trackerCompanionAutoLoopApplied` settings flag) applies `applyTrackerCompanionAutoLoopDefaults()`:

- `trigger: 'auto'` — the tracker runs on its own model/profile after **every** reply;
- `rawPrompt: true` — **new companion option**: the agent prompt is sent verbatim as the system message, with no appended format instructions (tracker prompts define their own exact output format and broke when wrapped). The short-lived tracker "state only" instruction from the previous commit is removed in favor of this. Editor checkbox: *Use agent prompt as-is*. Batch task blocks also skip the per-task format lines for raw-prompt agents;
- `feedback.enabled: true` — the latest state is injected back into the next main generation (placement still controlled by the agent's injection settings);
- `displayMode: 'panel'` — state lives in the slide-out Tracker panel, never as chat cards.

Net effect: main LLM writes the scene → tracker LLM updates state → state rides into the next main prompt, hands-free.

### Panel access from the chat screen

A map-location toggle button now sits in the send bar next to the wand-menu button (runtime-injected into `#leftSendForm`, visibility synced with the edge handle). Also fixed `readCompanionConfigFromEditor` clobbering `displayMode: 'panel'` back to `'card'` on every editor save.

### Theme/browser styling hardening

- Every `color-mix()` declaration now has a plain `var()`/rgba fallback declared before it — browsers without color-mix support (e.g. older iOS Safari) were dropping the declarations entirely, leaving cards borderless and untinted.
- `<summary>` disclosure markers hidden properly (`list-style: none` + `::marker { content: none }`, not just `::-webkit-details-marker`).
- Card/dashboard/panel action buttons get full `appearance`/`box-shadow`/`margin`/`font` resets, and the in-chat card actions gain a `#chat`-scoped high-specificity selector so theme rules like `#chat .mes button { … }` can no longer inflate them into giant default buttons.

### Tests

8 suites / **128 tests** green: auto-loop defaults (idempotent, tracker-companion-only, preserves feedback depth), raw-prompt verbatim system message vs. default format instruction, panel display mode surviving editor round-trips via the widened display-mode parsing, plus the existing panel/dashboard/cards coverage.

platberlitz:
## Follow-up 3 (`01c11e7b2`): tracker story guard + mobile panel fixes

- **Trackers were continuing the story after their state block** when running with raw prompts (prompts written for ride-along use read as “write a reply” when sent standalone). Tracker-category companions now always get one guard line appended after their verbatim prompt — *output only the tracker state, do not continue the story* — in both single and batch runs. `rawPrompt` still suppresses the generic per-format instructions; the guard is the only addition.
- **Send-bar panel button dead on mobile**: the mobile shell re-arranges `#leftSendForm`'s children, which can orphan a directly bound handler — the button now uses a delegated document-level click handler. The panel and edge handle also move from `z-index: 3000` (tying the shell's modal layer) to the shell's popout layer (`var(--sb-z-popout, 4000)`), so the panel can't slide out underneath mobile shell UI.
- Context scope note: the tracker's reading window is already the companion **Context messages** setting (1–50, editor field), sliced as the last N messages before the reply — verified wiring, no change needed.

8 suites / 129 tests green.

platberlitz:
## Follow-up 4 (`6c1d7c1b9`): the iOS "nothing happens" mystery was stale CSS

Diagnosis from an iOS Safari report (panel button doing nothing + the Companions dashboard rendering as unstyled centered text with giant default buttons, while the same build looked correct on desktop):

- Unversioned **JS** is served `Cache-Control: no-cache` (the middleware's own "stale mixed frontend graphs" rationale), but **CSS** fell into the `public, max-age=3600` bucket.
- The service worker doesn't intercept `/scripts/` paths, and the extension asset `?v=` query only changes per app release.
- Net result: a browser can legitimately pair fresh extension JS with an hour-stale `style.css`. The JS created the tracker panel and its send-bar button, but the stylesheet holding the panel's `position: fixed`/`transform` rules was old — so the panel "opened" as an invisible unstyled div, and every `.ica--cdash-*` dashboard rule was simply absent. This also retroactively explains the earlier "ugly, not theme-aware" card screenshots during this PR's iteration.

**Fix:** `src/middleware/frontend-assets.js` now puts `.css` in the same no-cache revalidation bucket as `.js`/`.mjs` (ETag/304 keeps repeat loads cheap); middleware test extended. Also added the `-webkit-backdrop-filter` prefix for Safari blur on the panel.

Note for reviewers: this is a server-wide caching policy change, deliberately small. Devices holding a stale copy right now will self-heal within the old hour TTL (or on one manual reload); after that, every CSS change is picked up on next load. 8 suites / 129 in-chat-agents tests green; the middleware test runs in CI (needs root deps).

platberlitz:
## Follow-up 5 (`1619af6ae`): QoL round from live mobile usage

- **Agents page tabs** — All / ⭐ Quick Access / Pre / Post / 🧑‍🚀 Companions strip above the filters; selection persists per device (localStorage). Quick Access shows only the pinned grid; phase tabs exclude companions; search/category filters stack on top.
- **Dedicated companion connection profile** — new settings select (`#ica--companionConnectionProfile`): companions resolve agent override → companion default → regular extension default chain (`resolveCompanionConnectionProfile`), so trackers/notes run on a cheap model while post-gen passes keep theirs. Batch grouping uses the same resolution.
- **Tracker panel QoL** — sections follow the agents-page order (`injection.order`, orphans last) instead of alphabetical-ish insertion order; the edge handle is **draggable** (pointer capture, 6px click-vs-drag threshold, clamped 8–92%, per-device persisted — also addresses it spawning awkwardly at the top on the mobile shell); **click-outside closes the panel** (250ms opening grace so the opening tap doesn't self-close).
- **Clickable choices** — `<li>` lines in companion bodies (cards and panel) insert into `#send_textarea` on click: list enumeration stripped (`1.`, `-`, `•`, `B)`…), replace-when-empty / append-on-new-line otherwise, native `input` event dispatched, panel closes itself afterwards. Built for CYOA Choices and Direction Menu, works for any list-style note.
- **Small-model output discipline** — reported on a smaller model: CYOA emitted a roleplay reply instead of choices. Every companion run now carries an anti-roleplay guard line (trackers keep the state-only wording; everything else gets *“you are an auxiliary companion, not the roleplayer”*) plus a `[Task]` anchor appended **after** the conversation context — the tail of the prompt ends with RP dialogue, which is exactly what small models latch onto. Applied to single and batch runs; `rawPrompt` still only suppresses format instructions.

8 suites / **134 tests** green.

platberlitz:
## Follow-up 6 (`HEAD`): iOS drag + tappable choices + execution mode split

- **Handle drag dead on iOS Safari** (incognito, so not cache): iOS ignores `touch-action` on `position: fixed` elements and converts the drag into a page scroll, firing `pointercancel`. Fixed with a non-passive `touchmove` `preventDefault()` on the handle plus `-webkit-touch-callout/user-select: none`.
- **CYOA choice clicks did nothing**: the bundled CYOA/Direction Menu templates ship no beautifier regex and their prompts emit `[CHOICES]` + numbered lines — with showdown's `simpleLineBreaks` that renders as one `<p>` with `<br>`s, never `<li>`, so the list-item click delegation never matched anything. New render-time `decorateChoiceLines()` pass wraps choice-looking content (list items, enumerated paragraphs, `<br>`-separated enumerated segments) in real `<button class="ica--choice-line">` elements — robust against output shape and natively tappable on iOS. Runs after sanitization; only our own button wrapper is added.
- **Separate Companion Execution dropdown** (parallel/sequential) — companions no longer piggyback on the append-agents execution mode; `globalSettings.companionExecutionMode` drives both auto and run-all paths.
- **Removed the send-bar map icon** at user request — redundant with the draggable edge handle (wand item and dashboard button remain).

8 suites / 134 tests green.

platberlitz:
## Follow-up 7 (`45fd79088`): companion context access + concurrency + mobile follow-through

- **Context access**: companions can now include the **Author's Note** (`chat_metadata.note_prompt`, global default fallback) and the **System Prompt** (`power_user.sysprompt.content`) in their calls, joining the existing persona / character card / World Info toggles — all editable per agent, and **all five default ON for every companion** per user request. Normalization applies new defaults only to absent keys (explicit opt-outs survive), and the auto-loop migration is now versioned (v2) so existing companions get the access grant once.
- **`Run Companions Alongside Post-Gen Passes`** toggle: starts the companion stage concurrently with post-generation passes instead of after them — faster turnarounds, with the documented tradeoff that companions see the pre-transform reply.
- **`Trackers → Companions`** toolbar button: one confirmed click converts every inline tracker to companion execution (auto loop + panel).
- Removed the Agents page intro blurb at user request.
- **Mobile handle, part 3**: v1 localStorage could hold a scroll-corrupted position from the era when iOS hijacked the drag — pinning the handle to the top edge with no way to drag it back (user screenshot confirmed it sitting at the 8% clamp). Storage key bumped to v2 (bad value abandoned, default restored) and `pointercancel` now *restores* the pre-drag position instead of persisting mid-hijack coordinates.
- **Choice taps**: document-level catch-all handler + toast feedback ('Added to the message box' / 'Could not find the message box') so remaining mobile reports can distinguish a dead tap from a mis-targeted textarea.

8 suites / **137 tests** green.

platberlitz:
## Follow-up 8 (`HEAD`): companions see their previous states by default

The `[Your previous notes]` context section (agent's last `historyDepth` done results, walked back from the current message) existed but defaulted **off** — and notably the tracker auto-loop never enabled it, so converted trackers were re-deriving state from scratch every turn instead of evolving their last state. `includeHistory` now defaults ON for all companions, joins the context-access defaults, and the migration version bumps to **3** so users already migrated under v2 (including the reporting user) get it applied once. Explicitly saved opt-outs survive normalization, and the editor checkbox (*Include prior notes*) still controls it per agent alongside *History depth* (1–10).

New runner test covers the section appearing with stored prior state and staying absent when opted out. 8 suites / **138 tests** green.

platberlitz:
## Follow-up 9 (`HEAD`): touch-native handle drag + choice detection in beautified markup

- **Handle drag, definitive version**: stopped patching iOS pointer-event quirks one at a time (ignored `touch-action` on fixed elements, scroll-takeover `pointercancel`, spotty `pointermove` after capture) — touch input now drives the drag through **raw touch events** with a non-passive `preventDefault` on every move of the gesture, which WebKit honors universally. Pointer events remain for mouse/pen only; threshold, clamping, click suppression, cancel-restore, and persistence are shared between paths.
- **Choice taps in beautified output**: user-installed CYOA/direction agents can carry beautifier regexes rendering choices as styled HTML rows, which the markdown-list/`<br>`-split detection never matched → zero buttons → dead taps. `decorateChoiceLines` gains a **deepest-match pass** wrapping any innermost element whose own text reads as a single enumerated choice, and the `<br>`-split pass now covers `div`s too.

8 suites / 138 tests green.

platberlitz:
## Follow-up 10 (`29a1c67a4`): pixel-anchored handle + four QoL fixes

- **Handle pinned to the top edge, immovable**: new diagnosis that finally fits every report — CSS percentage `top` on `position: fixed` resolves against whatever **containing block** the page creates (mobile shells can make body one), which pins the handle to the top no matter what gets stored or dragged. The handle is now placed via **inline pixel positions computed from the visual viewport** (stored fraction or middle edge), re-applied on resize/rotation, with drag math using `visualViewport.height`. Combined with the touch-native drag from `97102c366`.
- **Swipe-safe feedback**: feedback injection skips the chat tail when it's an assistant message — that generation is rewriting it (swipe/regenerate/continue), so its own stored companion state is stale and must not be fed back. Detection is tail-based because `normalizeGenerationType` collapses swipe/regenerate into 'normal'.
- **NPC Motivator v2** ships as an auto-loop companion (raw prompt, panel display, full context access, feedback depth 1, 4096 tokens) instead of a pre-generation intercept patch; existing installs see the standard template-update pill.
- **Phase round-trip**: converting to companion records the prior phase (`companion.inlinePhase`) and converting back restores it — previously every round-trip left agents as post-gen even if they were pre-gen.
- **Tracker panel**: header **Regenerate All** button (last reply, all companions) and a per-agent **wrench Fix** action that re-runs with a strict repair instruction — the companion-flavored 'Fix Trackers' for when smaller models roleplay instead of producing the required output.

8 suites / **141 tests** green.

platberlitz:
## Follow-up 11 (`HEAD`): drag confirmed working on iOS 🎉 — four-edge docking

User confirmed the touch-native drag works. Built on it per request:

- Handle renamed **Trackers → Companion** (astronaut icon), wand item and dashboard button now say *Companion Panel*, panel header says *Companions*.
- **Drag anywhere, dock to the nearest edge**: release snaps to left/right/top/bottom at the release point's fraction along that edge, persisted as `{ edge, fraction }` (legacy bare-number values still parse as right-edge docks). The handle re-orients per edge (vertical label + edge-rounded corners on the sides, horizontal on top/bottom, safe-area padding per edge).
- **The panel opens from the docked edge**: side drawer on left/right (min(420px, 94vw) wide), a `min(70dvh, 560px)` sheet sliding from top/bottom, each with matching transforms and borders via `data-edge` variants.

New pure-function coverage for `resolveHandleDock` (nearest-edge + fraction clamping) and `parseStoredHandlePosition` (legacy/JSON/garbage forms). 8 suites / **143 tests** green.

platberlitz:
## Follow-up 12 (`HEAD`): panel tap fix + unified guard wording

- **Panel auto-closing on taps**: the deepest-match choice detection from follow-up 9 was too greedy — bullet-prefixed tracker state lines matched as 'choices', so ordinary taps inserted text and triggered the close-after-insert behavior. Detection now requires a real enumerator (`1.` / `2)` / `B)`; markdown `<li>` keeps bullet support via the list pass), and panel inserts no longer auto-close — the toast is the confirmation.
- **Guard unified and reworded** (user request): every companion now gets a single leading guard — *“Pause all roleplay and output only the following.”* — ahead of the agent prompt in single runs and ahead of each batch task block. Replaces both the tracker-specific and generic guard variants.

8 suites / 143 tests green.

platberlitz:
## Follow-up 13 (`HEAD`): regression fix — panel could not stay open

Straight regression from the four-edge docking commit: `.ica--tpanel.is-open` (slide-in transform) and the per-edge closed transforms (`.ica--tpanel[data-edge="right"]` etc.) tie at (0,2,0) specificity, and the edge rules sit later in the stylesheet — so the closed `translateX(100%)` overrode the open state and the panel never stayed out (observed as slide-out-then-back / clicks doing nothing once caches settled). Open transforms are now per-edge `[data-edge].is-open` selectors at (0,3,0). Tests unaffected (CSS-only); 143 green.

platberlitz:
## Follow-up 14 (`HEAD`): bundled Companions group

Added `grp-companions` to `templates/groups.json` for the Agent Groups UI — installs all seven bundled companions in one click (continuity, relationship lens, director's commentary, actor interview, lorebook scout, memory shard, and NPC Motivator v2). All referenced template ids validated against the catalog; 143 tests green.

platberlitz:
## Follow-up 15 (`HEAD`): edit buttons in the companion panel

Each agent section in the slide-out panel now has an edit (pen) action next to regenerate/fix/jump — also shown on agents with no stored state yet, hidden for orphaned results whose agent was deleted. The editor opens through a new `configureCompanionPanel({ openEditor })` seam (mirroring the dashboard's, no index.js import cycle), with the panel closing first so the popup never fights the sheet edges. 143 tests green.

platberlitz:
## Follow-up 16 (`HEAD`): panel is the universal display default

User installed bundled companions (via the new Companions group) and their notes rendered as in-chat cards — the templates still shipped `displayMode: 'card'` and only trackers were forced to panel. Per the user's rule (*all companions strictly in the page*): all six companion templates flip to **panel** (v2, standard update pill), `createDefaultCompanionConfig` defaults to panel, and migration **v4** moves every existing card-mode companion over. In-chat cards remain as an explicit per-agent editor opt-in, and old stored card-mode results vanish from the chat immediately because the ledger consults the agent's current display mode. 8 suites / **144 tests** green.

platberlitz:
## Follow-up 17 (`HEAD`): Narration Voice commentary + Plot Compass

- **Director's Commentary v3** now speaks as the story's narrator stepping out from behind the page, styled by the user preset's active **Narration Voice**: the prompt consumes `{{getvar::narration}}` (companion prompts run the full macro engine, so the preset's `Voice: …` setvar entries flow straight in) and falls back to the voices' shared register — intimate, mischievous, conspiratorial direct address with razor-sharp asides — when no voice is set.
- **New bundled Plot Compass companion**: the user types a standing `OBJECTIVE:` line into its note via the new **edit-note** panel action; each auto run carries the objective forward verbatim and plans around it (progress, numbered next beats, obstacles, one tappable suggested move), with feedback enabled so the plan also steers the next main generation. Joined the Companions group.
- **Panel note editing**: per-agent file-pen action reusing the card editor popup (`editCompanionResult` now exported) — stored state is hand-editable directly from the panel.

8 suites / 144 tests green; group↔catalog ids validated.

platberlitz:
## Follow-up 18 (`HEAD`): memory shard threshold + compaction, panel edit semantics

- **`minContextTokens` companion option** (0–200k, editor field): auto runs skip until the chat's estimated token total reaches it (per-message token accounting with a chars/4 fallback); stateless agents below threshold stay out of the panel. **Memory Shard v3** flips to auto trigger at **30000 tokens** — the original Memory Sharding QR's '~30k context' guidance, automated.
- **Shard compaction**: once a shard exists, its panel section offers *Hide story above this shard* — confirm popup, then `hideChatMessageRange(0, shardIndex - 1)` excludes the summarized messages from prompts while the feedback-injected shard carries the history. Messages stay visible and unhideable.
- **Panel edit semantics** (user feedback): the pen icon now edits the **state text** (was: agent settings), agent settings moved to a gear icon, and the duplicate file-pen button is gone.

8 suites / **146 tests** green.

platberlitz:
## Consolidated Commit Summary

This consolidates the earlier PR follow-up comments into one commit-by-commit summary. It covers all 28 commits currently in the PR.

Overall scope:

- Adds Companion Agents as a native In-Chat Agents execution path: sidecar AI runs that can render notes, store panel state, feed prior notes back into the main prompt, and run independently from inline pre/post agents.
- Adds the companion runner, note-card UI, dashboard, slide-out panel, conversion flows, companion-specific settings, bundled templates, and broad unit coverage.
- Iterates heavily on tracker conversion, mobile/iOS behavior, stale CSS caching, small-model prompt discipline, panel-first display defaults, and memory-shard compaction.
- Current diff: 33 changed files, about +7543 / -285.

### 1. `6e12fefb68` - `feat(ica): add companion agents`

- Introduces the core `execution: 'inline' | 'companion'` axis in In-Chat Agents and adds companion configuration defaults in the store.
- Adds `companion/companion-runner.js` for auto/manual/keyword companion runs, regex-literal trigger support, batching with fallback to individual runs, cancellation, per-swipe result persistence, and optional feedback injection into later prompts.
- Adds `companion/companion-ui.js` for collapsible companion note cards under assistant replies, including regenerate/edit/copy/delete actions, status states, markdown/html/text formatting, and collapse persistence.
- Wires the editor/settings/index integration, initial UI styling, and the first bundled companion templates: Continuity Companion and Relationship Lens.
- Adds baseline tests for store normalization, generation UI wiring, runner behavior, template catalog sync, and pre-intercept summary interactions.

### 2. `b39dc56dd8` - `fix(ica): stop extra-only snapshot refreshes from reloading the chat`

- Fixes the history-wipe bug documented in the earlier PR comment: converting an inline regex/tracker agent to companion removed now-inapplicable regex snapshots from every message's `extra`, and each off-screen snapshot refresh fell back to `saveChat()` plus `reloadCurrentChat()`.
- The reload path briefly clears in-memory chat state with `clearChat({ clearData: true })`; raced/interrupted reloads could persist an emptied chat. Conversion made this deterministic, but the same class of bug also existed for disabling/deleting inline regex agents.
- Threads `skipReloadFallback` through `scheduleMessageRefresh()` into `refreshMessageAfterMutation()` for snapshot-only bookkeeping updates. Rendered messages still update in place; off-screen messages no longer force full chat reloads.
- Preserves the existing save/reload fallback for real text mutations and prevents a later bookkeeping refresh from downgrading a pending full refresh for the same message index.
- Adds regression tests for conversion safety, off-screen snapshot refresh behavior, and the non-downgrade case.

### 3. `cd6e9e930e` - `feat(ica): one-click inline/companion conversion on agent cards`

- Adds `convertAgentExecution(agent, target)` in the store.
- Inline to companion conversion preserves prompt, regex scripts, injection settings, conditions, and companion config, then normalizes companion execution.
- Companion to inline conversion keeps the companion config for round-tripping and moves `category: 'companion'` agents to `custom`, because category normalization otherwise re-derives companion execution.
- Adds per-agent card conversion actions and a bulk select-mode "To Companion" action.
- Reuses the safe snapshot refresh path fixed in the previous commit.

### 4. `c4c262b253` - `feat(ica): companion dashboard with wand menu and toolbar entry`

- Adds `companion/companion-dashboard.js`, opened from a wand-menu item and a Companions toolbar button.
- Dashboard rows show enable toggles, trigger/display/format/batch/feedback pills, run-on-last-reply, edit, and convert-to-inline actions.
- Adds a one-click conversion section for eligible inline agents, New Companion and AI Maker entry points, Run All on Last Reply, and a recent-notes browser with jump-to-message.
- Live-updates while open on companion result events, preserving scroll and removing listeners on close.
- Uses `configureCompanionDashboard()` hooks so dashboard behavior can call editor/toggle/conversion flows without importing `index.js` back and creating a cycle.

### 5. `b7b9eb8867` - `fix(ica): apply agent regex to companion cards and missing render events`

- Runs the owning agent's regex scripts over companion output before markdown/html/text formatting, preserving tracker beautifiers after conversion to companion execution.
- Keeps sanitization after the regex pass, so formatted companion cards still go through the safe rendering path.
- Re-renders companion cards on lazy history load, message edit, and message delete events.
- Re-renders all message cards after deletion because message indices shift.

### 6. `17f6f131ed` - `feat(ica): add directors-commentary, actor-interview, and lorebook-scout companion templates`

- Adds Director's Commentary as an auto, batched scene-commentary companion.
- Adds Actor Interview as a manual companion that can include the character card and conduct an in-character backstage interview.
- Adds Lorebook Scout as a manual companion that drafts one-concept lorebook entries with trigger keys and includes World Info so it can avoid duplicating already-covered lore.
- Syncs the source template files with `templates/index.json` and expands template tests.

### 7. `632cfef525` - `test(ica): cover companion trigger scanning and per-swipe note persistence`

- Adds coverage that keyword triggers scan the latest user message, including continue-generation cases where the chat tail is an assistant message.
- Verifies assistant-only keyword mentions do not trigger companions.
- Verifies regex-literal keyword triggers compile and match correctly.
- Locks in companion note mirroring into `swipe_info`, so notes survive swiping away and back.

### 8. `7134c65fa8` - `feat(ica): add memory-shard summary companion template`

- Adds Memory Shard, a token-lean companion version of the bundled Memory Sharding quick reply.
- Keeps the minimum reconstruction set: timeline, characters, relationships, weighted events, dialogue keys, threads, and current state.
- Drops the original long protocol/examples and uses terse pipe-delimited state formats.
- Runs over a 30-message context, builds on prior shards through note history, and feedback-injects the latest shard as rolling long-term memory.

### 9. `1a10303127` - `docs: record the index.html companion message button in the upstream touch ledger`

- Updates `docs/upstream-touch-ledger.md` for the static `public/index.html` companion message button row.
- Documents the upstream touch point while keeping the runtime logic inside the extension.

### 10. `30bdbabf0e` - `feat(ica): slide-out tracker panel and panel display mode for companions`

- Adds `displayMode: 'panel'` so companion results can be stored and used for feedback without rendering as chat cards.
- Introduces `companion/companion-panel.js`: a fixed slide-out panel showing each companion's latest rendered state, regenerate action, jump-to-source-message action, and previous-state history.
- Makes tracker-to-companion conversion default to panel display so tracker state leaves the chat flow.
- Adds panel entry points through an edge handle, wand-menu item, and dashboard button.
- Adds tracker-specific state-only prompting so standalone tracker companions stop wrapping state blocks in scene prose.
- Stores connection profile display names instead of raw ids and cleans UUID suffixes from companion card/panel headers.

### 11. `bc25e48873` - `feat(ica): automatic tracker loop, raw prompts, chat-bar panel button, and theme-safe styling`

- Makes converted trackers run in a full automatic loop by default: auto trigger after each reply, panel display, feedback enabled, and raw prompt mode.
- Adds `rawPrompt` so tracker prompts are sent verbatim instead of receiving generic companion format instructions that can distort tracker output formats.
- Adds a one-time migration for already-converted tracker companions so existing users get the same auto-loop/panel behavior.
- Adds a chat send-bar panel toggle and fixes the editor save path that could clobber `displayMode: 'panel'` back to `card`.
- Hardens styles with `color-mix()` fallbacks, better disclosure marker resets, and high-specificity/reset styling for action buttons so themes do not inflate companion controls.

### 12. `01c11e7b26` - `fix(ica): tracker story guard, delegated panel button, and shell-aware z-index`

- Restores a tracker guard after raw-prompt testing showed tracker prompts could still continue the story after emitting state.
- Applies the guard in both single and batch companion runs while keeping raw prompt mode free of the generic format instructions.
- Changes the send-bar panel button to delegated document-level click handling because the mobile shell can rearrange send form children.
- Moves the panel and handle to the shell popout z-index layer so the panel cannot slide out underneath mobile UI.

### 13. `6c1d7c1b9d` - `fix(server): revalidate unversioned stylesheets like scripts`

- Fixes the iOS "nothing happens" report traced in the earlier comments to stale CSS: extension JS was revalidating, but CSS could be cached for up to an hour.
- Puts unversioned `.css` assets in the same no-cache revalidation bucket as `.js`/`.mjs`, avoiding fresh JS paired with stale companion styles.
- Explains why the dashboard could appear as unstyled text and the panel could be invisible even though JS created it.
- Adds Safari `-webkit-backdrop-filter` support for the panel and extends middleware coverage.

### 14. `1619af6ae3` - `feat(ica): agent tabs, companion profile default, panel QoL, and small-model guards`

- Adds Agents page tabs: All, Quick Access, Pre, Post, and Companions, with per-device persistence.
- Adds a dedicated Companion Connection Profile setting, resolving agent override to companion default to regular extension default.
- Orders panel sections by agent order, makes the edge handle draggable with persisted/clamped placement, and closes the panel on outside click.
- Makes choice-like companion output insert into the message box when clicked, supporting CYOA/direction-menu use cases.
- Adds anti-roleplay guard wording and a `[Task]` anchor near the prompt tail so smaller models are less likely to continue the roleplay instead of doing the companion task.

### 15. `a7247e3540` - `fix(ica): iOS handle drag, tappable choice buttons, companion execution mode, drop chat-bar icon`

- Addresses iOS Safari ignoring `touch-action` on fixed elements by adding non-passive touch handling and suppressing long-press selection/callout on the handle.
- Adds `decorateChoiceLines()` so list items, enumerated paragraphs, and `<br>`-separated enumerated lines become real tappable buttons.
- Adds a separate companion execution mode setting, so companions can run parallel or sequential independently from append-agent execution mode.
- Removes the redundant send-bar map icon after the draggable edge handle became the preferred access point.

### 16. `45fd79088d` - `feat(ica): companion context access, concurrent stage, bulk tracker conversion, and mobile fixes`

- Adds companion access toggles for Author's Note and System Prompt, alongside persona, character card, and World Info.
- Defaults all five context toggles on for companions while preserving explicit opt-outs through normalization.
- Adds a versioned migration so existing companions receive the new context defaults once.
- Adds "Run Companions Alongside Post-Gen Passes" to start companion work concurrently with post-generation passes, with the tradeoff that companions see the pre-transform reply.
- Adds a "Trackers -> Companions" toolbar button for confirmed bulk conversion of inline trackers.
- Bumps mobile handle position storage to avoid old scroll-corrupted iOS positions, restores starting position on cancelled drags, and adds document-level choice tap fallback/toasts.

### 17. `e2c28279e2` - `feat(ica): companions see their previous states by default`

- Defaults `includeHistory` on for companions so they receive their own recent prior notes in the `[Your previous notes]` context section.
- Fixes the tracker auto-loop behavior where trackers were re-deriving state from scratch instead of evolving their previous state.
- Bumps the migration version so users already migrated under the previous context-access version also receive prior-note defaults.
- Preserves explicit opt-outs and adds runner coverage.

### 18. `97102c3666` - `fix(ica): drive the handle drag with raw touch events on touch devices`

- Replaces the earlier iOS pointer-event workaround with raw touch events for touch devices.
- Uses non-passive `preventDefault()` during touch drag moves, which iOS WebKit reliably honors.
- Keeps pointer events for mouse/pen and shares begin/move/finish logic across both paths.
- Preserves threshold, clamping, click suppression, cancel restore, and per-device persistence behavior.

### 19. `a08be0037d` - `fix(ica): detect choice lines inside beautified companion markup`

- Fixes installed CYOA/direction agents whose beautifier regexes turn choices into styled HTML rows instead of plain markdown or `<br>` lines.
- Adds a deepest-match pass that wraps innermost elements whose own text is one enumerated choice.
- Extends the `<br>` split pass to cover `div` containers as well as paragraphs.

### 20. `29a1c67a45` - `fix(ica): viewport-pixel handle placement, swipe-safe feedback, NPC Motivator companion, phase round-trip, panel fix buttons`

- Fixes handle placement by using inline pixel positions computed from the visual viewport instead of CSS percentages on fixed elements, which mobile shells could resolve against unexpected containing blocks.
- Re-places the handle on viewport resize/rotation and uses visual viewport dimensions for drag math.
- Makes feedback injection swipe-safe by skipping the assistant tail when that generation is being rewritten by swipe/regenerate/continue.
- Converts bundled NPC Motivator v2 into an auto-loop companion with raw prompt, panel display, full context access, feedback, and 4096 max tokens.
- Records the prior inline phase in `companion.inlinePhase` and restores it when converting a companion back to inline.
- Adds panel Regenerate All and per-agent Fix actions for repairing malformed companion/tracker output.

### 21. `43676bc0f2` - `feat(ica): four-edge docking for the companion handle and panel`

- Renames the handle from Trackers to Companion and lets it dock to the nearest viewport edge: left, right, top, or bottom.
- Persists placement as `{ edge, fraction }`, while still parsing the legacy bare-number right-edge value.
- Re-orients the handle per edge, with vertical labels on sides and horizontal labels on top/bottom.
- Makes the panel open from the docked edge as a side drawer or top/bottom sheet, with edge-specific transforms, borders, and safe-area padding.
- Adds pure-function coverage for docking and stored-position parsing.

### 22. `adb945de9a` - `fix(ica): keep the panel open on choice taps and unify the leading guard`

- Fixes over-greedy choice detection that treated bullet-prefixed tracker state lines as tappable choices.
- Requires a real enumerator for deepest-match choice detection while keeping markdown list handling in the list-item pass.
- Stops panel choice inserts from auto-closing the panel; the toast is now the confirmation.
- Unifies companion guard wording to the leading line: `Pause all roleplay and output only the following.`
- Applies that guard to every companion in single runs and batch task blocks.

### 23. `d4228edc31` - `fix(ica): panel stayed off-screen - open transform lost to edge rules`

- Fixes a CSS specificity regression from four-edge docking.
- The generic open selector tied with later per-edge closed transforms, so the closed translate could win and keep the panel off-screen.
- Uses per-edge open selectors such as `.ica--tpanel[data-edge=...].is-open` so the open transform outranks closed edge rules.

### 24. `044251aa23` - `feat(ica): bundle a Companions group for the agent groups UI`

- Adds `grp-companions` to `templates/groups.json`.
- Lets users install the bundled companion set in one click through the Agent Groups UI.
- Initial group includes Continuity, Relationship Lens, Director's Commentary, Actor Interview, Lorebook Scout, Memory Shard, and NPC Motivator v2.

### 25. `2c1f76a6db` - `feat(ica): per-agent edit button in the companion panel`

- Adds an edit action to each companion panel section for agents that still exist, including agents with no stored state yet.
- Hides edit for orphaned stored results whose agent has been deleted.
- Adds `configureCompanionPanel()` hooks, mirroring the dashboard seam, so the panel can open the agent editor without importing `index.js` and creating a cycle.
- Closes the panel before opening the editor popup.

### 26. `49cb7a303e` - `fix(ica): companions live strictly in the panel by default`

- Fixes bundled companions still installing with `displayMode: 'card'`, which made new companion notes render inside chat.
- Changes bundled companion templates to panel mode and bumps their versions for standard update-pill behavior.
- Changes the default companion config to panel mode, while keeping in-chat cards available as an explicit editor opt-in.
- Adds migration v4 to move existing card-mode companions into the panel.
- Stored card-mode results disappear from chat immediately after the agent mode changes because rendering consults the current agent display mode.

### 27. `b3142f175f` - `feat(ica): narration-voice commentary, Plot Compass objective companion, panel note editing`

- Updates Director's Commentary v3 to consume the active Narration Voice through `{{getvar::narration}}`, relying on companion prompts running the full macro engine.
- Adds a fallback narrator register when no narration voice is set.
- Adds the Plot Compass companion: an objective-driven auto companion that carries a standing `OBJECTIVE:` forward, reports progress, plans next beats, identifies obstacles, and emits one tappable suggested move.
- Adds Plot Compass to the Companions group.
- Adds a panel action for editing stored companion note/state content from the panel, reused from the card editor flow.

### 28. `799c80803d` - `feat(ica): memory shard auto-runs at ~30k tokens with history compaction; panel pen edits content`

- Adds `minContextTokens` to companion config and editor UI, clamped from 0 to 200k tokens.
- Auto companions below their threshold skip runs, and stateless below-threshold companions stay out of the panel.
- Updates Memory Shard v3 to auto-run at about 30k estimated context tokens, matching the original quick-reply guidance.
- Adds a Memory Shard panel action to hide story above the shard through `hideChatMessageRange(0, shardIndex - 1)` after confirmation, letting the injected shard carry summarized history while messages remain visible.
- Refines panel edit semantics from the prior commit: the pen edits state text, agent settings move to a gear icon, and the duplicate file-pen action is removed.

## Test Signal From The PR Comments

- Latest reported focused run: `cd tests && npm run test:unit -- in-chat-agents` with 8 suites and 146 tests green.
- Earlier comments note root-suite failures in a fresh worktree were environmental missing-dependency issues, with CI expected to cover those suites.

platberlitz:
## Changelog-Ready Consolidated Summary

This is the same PR scope as the earlier consolidated commit summary, reorganized into release-note/changelog sections.

## Added

- Added **Companion Agents** as a native In-Chat Agents execution mode via `execution: 'inline' | 'companion'`.
- Added a companion runner that supports auto, manual, keyword, and regex-literal triggers.
- Added companion request batching with individual-run fallback, cancellation, per-swipe result persistence, and optional feedback injection into later main prompts.
- Added companion note rendering with collapsible cards, regenerate, edit, copy, delete, status states, persisted collapse state, and markdown/html/text display modes.
- Added a **Companions Dashboard** from the wand menu and extension toolbar with companion toggles, run/edit/convert actions, run-all, inline-agent conversion, New Companion, AI Maker, and recent-note browsing with jump-to-message.
- Added one-click conversion between inline and companion execution on agent cards, plus a bulk **To Companion** action in select mode.
- Added a **Companion Panel** slide-out/sheet UI for companion state, with latest state, previous state history, regenerate, regenerate-all, fix, jump-to-source, edit-state, and edit-agent actions.
- Added `displayMode: 'panel'` so companion output can live outside the chat while still being stored and available for feedback injection.
- Added four-edge docking for the companion handle: left, right, top, or bottom, with per-edge panel opening behavior.
- Added draggable, persisted companion handle placement with touch-native iOS drag handling.
- Added tappable companion choice lines that insert into the message box, including markdown lists, enumerated paragraphs, `<br>`-split lines, and beautified HTML choice rows.
- Added a dedicated Companion Connection Profile setting so companions can use cheaper/different models than normal pre/post agents.
- Added a separate Companion Execution Mode setting for parallel/sequential companion runs independent from append-agent execution mode.
- Added companion context toggles for persona, character card, World Info, Author's Note, System Prompt, and prior companion notes.
- Added `rawPrompt` companion mode for prompts that define their own output format, especially trackers.
- Added `minContextTokens` companion thresholding so auto companions can wait until the chat reaches a configured estimated token count.
- Added concurrent companion execution alongside post-generation passes, with companions seeing the pre-transform reply.
- Added a bulk **Trackers -> Companions** toolbar action.
- Added panel history compaction for Memory Shard: **Hide story above this shard** excludes summarized earlier messages from future prompts while keeping them visible.

## Templates

- Added **Continuity Companion** for automatic continuity notes and feedback.
- Added **Relationship Lens** for manual relationship/scene analysis.
- Added **Director's Commentary** for automatic batched commentary, later updated to use the active Narration Voice via `{{getvar::narration}}`.
- Added **Actor Interview** for manual in-character backstage interviews with character-card access.
- Added **Lorebook Scout** for one-concept lorebook draft entries with trigger keys and World Info dedupe context.
- Added **Memory Shard** as a token-lean rolling memory companion, later updated to auto-run around 30k context tokens and support history compaction.
- Converted **NPC Motivator v2** from a pre-generation intercept patch into an auto-loop panel companion with raw prompt, full context access, feedback, and 4096 max tokens.
- Added **Plot Compass** for an objective-driven rolling plan with progress, next beats, obstacles, and one tappable suggested move.
- Added a **Companions** agent group for one-click installation of the bundled companion set.

## Changed

- Changed tracker-to-companion conversion to default into an automatic panel loop: auto trigger, raw prompt, feedback enabled, and panel display.
- Changed companion defaults to panel-first behavior. In-chat cards remain available only as an explicit per-agent opt-in.
- Changed all companion templates to ship in panel mode by default.
- Changed companion context defaults so new companions can see persona, character card, World Info, Author's Note, System Prompt, and prior notes by default.
- Changed companion feedback behavior to skip stale assistant-tail state during swipe/regenerate/continue flows.
- Changed inline conversion round-tripping to preserve the original inline phase through `companion.inlinePhase`.
- Changed companion result rendering to store connection profile display names instead of raw profile ids.
- Changed companion card/panel headers to strip trailing UUID-like suffixes from agent names.
- Changed Agents page organization with All, Quick Access, Pre, Post, and Companions tabs.
- Changed panel section ordering to follow agent order, with orphaned stored results placed last.
- Changed panel editing semantics: the pen edits stored state text, while agent settings moved to a gear icon.
- Changed the companion guard instruction to a unified leading line: `Pause all roleplay and output only the following.`

## Fixed

- Fixed a severe chat-history wipe bug during regex snapshot refreshes. Snapshot-only `extra` mutations no longer trigger full `saveChat()` plus `reloadCurrentChat()` fallbacks for off-screen messages.
- Fixed snapshot refresh dedupe so bookkeeping-only refreshes cannot downgrade pending real text-mutation refreshes.
- Fixed companion card parity by applying the owning agent's regex scripts before markdown/html/text formatting, preserving tracker beautifiers after conversion.
- Fixed companion cards missing from lazy-loaded history and stale after message edits/deletes.
- Fixed tracker companions writing story prose around state blocks by adding explicit companion guard instructions.
- Fixed editor saves clobbering panel display mode back to cards.
- Fixed mobile shell panel button behavior by using delegated click handling and shell-aware z-indexing.
- Fixed stale CSS on iOS/Safari and other browsers by revalidating unversioned stylesheets like scripts instead of caching them for up to an hour.
- Fixed Safari panel blur by adding `-webkit-backdrop-filter`.
- Fixed theme compatibility with `color-mix()` fallbacks, disclosure-marker resets, and hardened action-button styles.
- Fixed iOS handle dragging by moving from pointer-event workarounds to raw touch events for touch devices.
- Fixed stored mobile handle positions corrupted by earlier iOS scroll hijacking by bumping storage keys and restoring pre-drag positions on cancellation.
- Fixed handle placement on mobile shells by anchoring with viewport-computed pixel positions instead of fixed-element CSS percentages.
- Fixed panel open/closed transform specificity after four-edge docking so open edge panels no longer stay off-screen.
- Fixed choice taps for CYOA/direction-menu output rendered as `<br>` lines or beautified HTML rows.
- Fixed over-greedy choice detection that treated bullet-prefixed tracker state as selectable choices.
- Fixed panel auto-closing on choice insertion; toast feedback now confirms insertion instead.
- Fixed companions re-deriving state from scratch by default by enabling prior-note history for companions.
- Fixed companion phase round-trip so pre-generation inline agents return to pre-generation after companion conversion and back.

## Migrations And Compatibility

- Added one-time tracker auto-loop migration for existing tracker companions.
- Added versioned companion context-access migrations so existing companions receive new context defaults once while preserving explicit opt-outs.
- Added migration v3 to enable prior-note history for already-migrated companions.
- Added migration v4 to move existing card-mode companions into the panel by default.
- Preserved legacy companion handle position parsing for older right-edge numeric values.
- Kept in-chat companion cards available as an explicit editor opt-in for users who want that display mode.
- Kept root behavior for true text mutations unchanged: real off-screen text changes still use the save/reload fallback.

## Docs And Tests

- Updated `docs/upstream-touch-ledger.md` for the static companion message button touch point in `public/index.html`.
- Added/expanded unit coverage for companion store normalization, conversion round-trips, tracker auto-loop defaults, context defaults, display-mode migrations, connection profile resolution, and list-tab filtering.
- Added/expanded runner coverage for trigger scanning, regex-literal keywords, per-swipe persistence, raw prompts, guard prompts, prior-note history, feedback behavior, token thresholds, and history compaction.
- Added companion dashboard coverage for partitioning, actions, live update listener lifecycle, and UI hooks.
- Added companion panel coverage for rendering, handle placement/docking parsing, panel actions, choice insertion, memory-shard compaction, and edit/fix controls.
- Added companion UI coverage for regex application, lazy-load/edit/delete re-render events, choice decoration, hidden/panel ledger exclusion, and name cleanup.
- Added template coverage for catalog/source sync, bundled companion templates, NPC Motivator v2, Plot Compass, and the Companions group.
- Latest focused test signal from PR comments: `cd tests && npm run test:unit -- in-chat-agents` passed with **8 suites / 146 tests**.

## Commit Traceability

- Core companion infrastructure: `6e12fefb68`
- Snapshot/history-wipe fix: `b39dc56dd8`
- Conversion actions: `cd6e9e930e`
- Dashboard: `c4c262b253`
- Card regex/render events: `b7b9eb8867`
- Initial companion templates: `17f6f131ed`, `7134c65fa8`
- Trigger/swipe tests and docs: `632cfef525`, `1a10303127`
- Panel and tracker loop: `30bdbabf0e`, `bc25e48873`, `01c11e7b26`
- CSS caching and theme/mobile fixes: `6c1d7c1b9d`, `a7247e3540`, `97102c3666`, `a08be0037d`, `d4228edc31`
- Companion settings/context/defaults: `1619af6ae3`, `45fd79088d`, `e2c28279e2`, `49cb7a303e`
- Panel docking/QoL: `29a1c67a45`, `43676bc0f2`, `adb945de9a`, `2c1f76a6db`, `799c80803d`
- Bundled groups/templates/Plot Compass: `044251aa23`, `b3142f175f`

platberlitz:
## Plain-English Summary

This PR adds **Companion Agents**: helper AIs that work alongside the main chat instead of writing directly into the story. They can track details, make notes, suggest next steps, summarize long chats, and help keep roleplay consistent without interrupting the main conversation.

## What Users Get

- A new **Companion Panel** that keeps helper notes and tracker state out of the chat itself.
- A **Companions Dashboard** where users can turn companions on/off, run them, edit them, convert existing agents, and review recent notes.
- One-click conversion for existing agents, especially trackers, so they can run quietly in the background instead of posting into the chat.
- Automatic tracker updates after replies, with tracker state fed back into future prompts so the main model remembers important details.
- Companion notes that can be edited, regenerated, fixed, copied, deleted, or used as clickable suggestions.
- A draggable Companion handle that works on desktop and mobile, including iOS Safari, and can dock to any screen edge.

## New Built-In Companions

- **Continuity Companion** keeps track of scene consistency.
- **Relationship Lens** summarizes character dynamics.
- **Director's Commentary** adds narrator-style observations.
- **Actor Interview** lets users ask character-focused backstage questions.
- **Lorebook Scout** suggests lorebook entries from new story facts.
- **Memory Shard** summarizes long chats and can hide already-summarized history from future prompts.
- **NPC Motivator** now runs as a background companion instead of patching prompts directly.
- **Plot Compass** helps carry a user objective forward and suggests next story moves.

## Why It Matters

- Long chats should be easier to manage.
- Trackers should no longer clutter the chat with state blocks.
- Smaller or cheaper models can handle background helper work.
- Mobile users get a cleaner, more reliable panel experience.
- Existing tracker setups can be upgraded with minimal manual work.
- Companion output can guide future generations without becoming part of the story text.

## Important Fixes Included

- Fixed a serious issue where converting or changing regex-based agents could wipe long chat histories.
- Fixed stale stylesheet caching that made the new UI appear broken or invisible on some browsers, especially iOS Safari.
- Fixed multiple mobile interaction problems around dragging the panel handle and tapping companion suggestions.
- Fixed companion/tracker prompts so helper AIs are less likely to accidentally continue the roleplay instead of doing their assigned task.

## Testing

The focused In-Chat Agents test suite passed with **8 suites / 146 tests** after these changes.

platberlitz:
## Update: Chatroom Companion + Panel Choice Auto-Close

Pushed `cd87994d8 feat(ica): add chatroom companion and close panel choices`.

This adds:
- Companion Panel now closes automatically after a CYOA/direction-menu choice is successfully inserted into the chat box, so mobile users are taken back to the conversation flow immediately.
- New bundled `Chatroom` companion template, inspired by EchoChamber, that creates a styled live audience reaction feed in the Companion Panel.
- Chatroom supports `mixed`, `in-world`, `discord/twitch`, `twitter/x`, `ao3/wattpad`, `newsroom`, and `thread-board/4chan` styles.
- NSFW EchoChamber styles/personas are deliberately excluded; adult source scenes are handled as non-explicit audience reactions.
- Chatroom uses safe inline HTML only and keeps feedback disabled, so it reacts beside the story without steering future generations.

Tests passed:
- `npm run test:unit -- in-chat-agents-companion-panel in-chat-agents-templates`
- `npm run test:unit -- in-chat-agents`

PR #457, opened 2026-06-12
fix: route initial load scroll through state machine

TheLonelyDevil9:
## What?
Routes the guarded initial-load bottom landing in `public/script.js` through the chat scroll state machine instead of resolving the scroll action directly.

## Why?
Initial chat load is the remaining scroll path called out in the mobile refactor plan after #454. Keeping it behind the existing rollout gate moves another upstream-origin scroll decision onto the shared `chat-render-lifecycle` seam while preserving the legacy fallback.

## How?
`scrollLoadedChatToBottomThroughLifecycle()` now resolves `CHAT_SCROLL_INTENT.INITIAL_LOAD` with `resolveChatScrollStateTransition()`, stores the returned `chatScrollState`, then applies the transition action with the existing bottom-scroll helper. Wiring and state-machine tests pin that the initial-load route uses the state seam and always lands in `pinned-bottom` with a forced bottom action. The upstream touch ledger records the updated adapter shape.

## Testing?
- `git diff --check`
- `node --check public/script.js`
- `npm run test:unit --prefix tests -- chat-render-lifecycle-scroll-state.test.js chat-render-lifecycle-script-wiring.test.js` (42/42)
- `npm run lint -- --quiet`
- `SILLYBUNNY_TEST_BASE_URL=http://127.0.0.1:4568 npm run test:e2e --prefix tests -- --browser=chromium chat-scroll-state-machine.e2e.js` (1/1)

The broader `chat-scroll-regressions.e2e.js -g "long chat initial render lands near the latest message"` was attempted with Chromium on `http://127.0.0.1:4568` but stopped at the existing sample-character setup blocker before exercising this change.

## Screenshots (optional)
Not applicable; this is a guarded scroll adapter change with Chromium runtime coverage rather than a visual UI change.

## Anything Else?
Targets `TLD/mobile-refactor`. The temporary test server used port 4568, and it was stopped after verification.

### DIFF
--- a/docs/upstream-touch-ledger.md
+++ b/docs/upstream-touch-ledger.md
@@ -30,11 +30,11 @@ This ledger tracks intentional SillyBunny divergence in upstream-origin files. I
 | Area | Chat lifecycle. |
 | Divergence reason | SillyBunny adds mobile batching, bottom pinning, manual-scroll suppression, long-chat anchoring, streaming DOM throttles, and issue `#167` scroll stability behavior. |
 | Target seam | `public/scripts/chat-render-lifecycle/`. |
-| Adapter shape | Keep exported compatibility functions in `public/script.js`; delegate bottom-scroll decisions, scheduling, scroll intent/state, anchor preservation, replacement/media anchoring, and update batching to the lifecycle module. |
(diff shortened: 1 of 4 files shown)


TheLonelyDevil9 (commented):
No blocking findings in the PR 4.4 review. Scope matches the initial-load scroll-state route in the refactor plan; checks are green for metadata, merge conflicts, ESLint, unit tests, frontend assets, and Bun runtime.

PR #504, opened 2026-06-18
fix: harden model sampling profile detection

platberlitz:
## Summary

Hardens the sampling preset system so it detects models by family rather than exact name, especially for OpenAI Custom models. This allows OpenAI Custom endpoints using OpenAI-family model IDs to share the same sampling profile as native OpenAI.

## Changes

- **Canonical profile keys**: `buildChatCompletionSamplingProfileKey()` normalizes OpenAI-family model IDs so `openai:gpt-4o`, `custom:openai/gpt-4o`, and `custom:chatgpt-4o-latest` all map to the same canonical key `openai:gpt-4o`
- **Legacy fallback**: `getChatCompletionSamplingProfileLookupKeys()` includes the old exact key `{source}:{model}` so previously saved profiles without canonical normalization continue to work
- **Automatic migration**: When a legacy profile is loaded, it is copied to the canonical key and saved via `saveSettingsDebounced`, so aliases share it going forward
- **Save deletes stale aliases**: Saving a profile writes the canonical key and removes duplicate legacy keys for the current model
- **Model family detection**: Recognizes OpenAI models under `custom` source via provider-segment detection and model ID pattern matching for `gpt-*`, `codex-*`, `omni-*`, `o1/o3/o4-*`
- **Unknown custom models stay isolated**: Non-OpenAI custom models get their own normalized custom keys (e.g. `custom:my-local-model`)
- Date-suffix stripping handles `o1-preview-2024-08-01` style versioned IDs

## Testing

Added 9 new test cases in `openai-preset-utils.test.js` covering:
- Canonical keys for native OpenAI models (openai, openai_responses, azure_openai)
- Shared canonical keys for OpenAI Custom models
- Separator and case normalization
- Unknown custom model isolation
- Legacy exact-key lookup and deduplication
- Null/empty source or model handling

All 33 sampling profile tests pass.

## Files changed

- `public/scripts/openai-preset-utils.js` — new `buildChatCompletionSamplingProfileKey` and `getChatCompletionSamplingProfileLookupKeys` exports
- `public/scripts/openai.js` — wired save/clear/apply to use canonical lookup with legacy migration
- `tests/openai-preset-utils.test.js` — new sampling profile key tests

### DIFF
--- a/public/scripts/openai-preset-utils.js
+++ b/public/scripts/openai-preset-utils.js
@@ -80,6 +80,115 @@ export function buildChatCompletionSamplingSettingsSnapshot(settings, settingsMa
     return structuredClone(snapshot);
 }
 
+const OPENAI_SAMPLING_PROFILE_SOURCE = 'openai';
+const OPENAI_SAMPLING_PROFILE_SOURCES = new Set([
+    'openai',
+    'openai_responses',
+    'azure_openai',
+]);
+const OPENAI_MODEL_PROVIDER_SEGMENTS = new Set([
+    'openai',
+    'openai-responses',
+    'azure-openai',
+]);
+
+function normalizeSamplingProfileSource(source) {
+    return String(source ?? '').trim().toLowerCase();
+}
+
+function normalizeSamplingProfileModel(model) {
+    return String(model ?? '')
+        .trim()
+        .toLowerCase()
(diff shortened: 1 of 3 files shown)


platberlitz:
Not quite right

PR #572, opened 2026-06-20
feat: Conversation mode & general UI polish

Geechan:
This PR changes the following for a polish pass on Conversation mode and the UI:

- Changes the header bars for all menu navigation to be consistent with each other.
- Cleans up LLMisms found in various UI subtext surrounding Conversation mode and the menu navigation.
- Updated the homescreen action card to open the Character menu instead of Conversation.
- Removes the Conversation sub-tab in favour of a toggle that switches between roleplay/conversation modes in the character menu. Easier to access this way.
- Conversation mode slightly reworked so only DMs that have messages in them will appear in the sidebar.
- Conversation mode now properly remembers the last chosen character when switching between roleplay/conversation mode.

### DIFF
--- a/public/css/sillybunny-conversation.css
+++ b/public/css/sillybunny-conversation.css
@@ -1995,6 +1995,91 @@
 }
 
 /* Style Add DM picker popover options */
+.sb-conversation-add-dm-title {
+    display: block;
+    font-size: var(--sb-type-meta);
+    font-weight: var(--sb-weight-title);
+}
+
+.sb-conversation-add-dm-description {
+    margin-top: 4px;
+}
+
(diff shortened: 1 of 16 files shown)


Geechan:
@platberlitz @TheLonelyDevil9 From now on, we really need to make absolutely sure we peer check every PR going forward, unless they're small bug fixes or documentation changes.

This is to:

1) Have more eyes and testing done on new code to prevent malware or other broken artifacts.
2) Help prevent conflicts with other PRs if one maintainer is working on something while the other is working on something else.

I will update the documentation later to reflect this, but let's make it an official policy.

PR #578, opened 2026-06-21
feat: make top bar title tappable to cycle label parts

platberlitz:
## Summary
- make the top bar title tappable to transiently preview configured, context size, character name, and custom text labels
- keep saved topbar label settings unchanged while previewing and reset previews on chat/context changes or after 5 seconds
- add accessible pointer/keyboard affordances and static unit coverage
- In essence: now you can just tap the top bar to go through Context Size, Character Name, or Custom Text (if applicable) in order to preserve top bar cleanliness.

## Testing
- npm run lint
- npm run build:frontend

### DIFF
--- a/public/css/sillybunny-tabs.css
+++ b/public/css/sillybunny-tabs.css
@@ -1184,10 +1184,33 @@ body:not(.movingUI) .drawer-content.maximized {
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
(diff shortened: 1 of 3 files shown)


Geechan:
This PR breaks desktop padding entirely. It appears to be based on the mobile refactor and is thus causing compatibility issues that are similar to those observed with the refactor. See the following screenshot: 
image

Closing for now.

PR #579, opened 2026-06-21
fix: prevent companion panel blur flash on close

platberlitz:
## Summary
- keep the companion tracker panel's per-edge off-screen transforms active while closed
- delay the closed-state visibility hide until the slide-out transition finishes so the backdrop blur leaves the viewport instead of flashing as a fullscreen slab
- Closing the Companion page added a full-screen blur for a second, this fixes that.

## Test plan
- NODE_OPTIONS=--experimental-vm-modules npx jest --config tests/jest.config.json tests/in-chat-agents-companion-panel.test.js tests/mobile-css-budgets.test.js --runInBand
- npm run lint


Geechan (approved):
Simple change; safe to merge.

PR #580, opened 2026-06-21
feat: improve accent profile settings

platberlitz:
## Summary
- make Accent Profiles in Appearance settings a collapsible inline drawer with persisted expanded/collapsed state
- keep Save Current visible while profiles are collapsed
- bump seeded accent profiles to 30 total and migrate existing users via the seed version
- More accent profiles because some of them were samey. Added some esoteric ones for the less creative and the lazy (like me)

## Tests
- npm run lint

### DIFF
--- a/public/css/sillybunny-theme.css
+++ b/public/css/sillybunny-theme.css
@@ -4643,6 +4643,20 @@ input[type='range']::-moz-range-thumb {
     gap: 12px;
 }
 
+.sb-accent-profiles-toggle {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: 10px;
+    flex: 1 1 auto;
+    min-width: 0;
+    cursor: pointer;
+}
+
+.sb-accent-profiles-toggle .inline-drawer-icon {
+    flex: 0 0 auto;
+}
+
 .sb-accent-profiles-copy {
     display: flex;
     flex-direction: column;
@@ -4665,6 +4679,10 @@ input[type='range']::-moz-range-thumb {
     min-height: var(--sb-control-min-height);
     flex: 0 0 auto;
     gap: 8px;
+}
+
+.sb-accent-profiles-content {
+    min-width: 0;
 }
 
 .sb-accent-profiles-list {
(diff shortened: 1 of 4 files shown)


Geechan:
Unrelated to this PR, but @platberlitz we're planning on making all PR descriptions manually written from now on. This is to:

1) Better describe what the actual intention behind a PR is, rather than having a model interpret it, leaving less room for error.
2) Reduce model verbosity that clutters up PR tracking and introduces completely unnecessary information.
3) Make it easier to backtrack on a PR and pinpoint exactly what causes a regression. 

It's more effort to do this, but it also encourages actually understanding what a PR does before submitting, too. The latest mobile refactor changes have caused many regressions, some of which are very difficult to pinpoint due to LLM wording. This issue has encouraged this change.

platberlitz:
> Unrelated to this PR, but @platberlitz we're planning on making all PR descriptions manually written from now on. This is to:
> 
>     1. Better describe what the actual intention behind a PR is, rather than having a model interpret it, leaving less room for error.
> 
>     2. Reduce model verbosity that clutters up PR tracking and introduces completely unnecessary information.
> 
>     3. Make it easier to backtrack on a PR and pinpoint exactly what causes a regression.
> 
> 
> It's more effort to do this, but it also encourages actually understanding what a PR does before submitting, too. The latest mobile refactor changes have caused many regressions, some of which are very difficult to pinpoint due to LLM wording. This issue has encouraged this change.

Okie. I will rewrite the current ones after my round of QoL PRs and bug fixes.

Geechan:
> make Accent Profiles in Appearance settings a collapsible inline drawer with persisted expanded/collapsed state

This PR seems to accidentally remove accent profiles entirely instead of reworking them to be in an in-line drawer. You'll want to double check this.

image

platberlitz:
Motherf

platberlitz:
Reworking this as well to match staging

platberlitz:
Still not here, thanks GLM :sob:

platberlitz:
Ok definitely need to redo it

PR #582, opened 2026-06-21
fix: stop custom endpoint Connect from clobbering profile keys

platberlitz:
## Summary
- The bug was essentially: It would keep cloning the other custom endpoint profile key, and then it refused to be used on the apikey field. So you'd (I would) have to use the Authorization: Bearer (key) header. This fixes that.
- Prevent Custom Endpoint Profile Connect from writing a new active CUSTOM secret when the selected profile already has a bound secret id.
- Clear the visible Custom key field and show `(saved secret)` for bound profiles so stale plaintext copies are not replayed.
- Keep legacy/keyless profile behavior by restoring the key input when no secret id is bound.

## Testing
- `npm run test:unit -- openai-proxy-preset-wiring.test.js`
- `npm run lint`
- `npm run test:unit`
- `npm run check:frontend-budgets && npm run build:frontend`
- `bun src/server-init.js`

### DIFF
--- a/public/scripts/openai.js
+++ b/public/scripts/openai.js
@@ -8673,7 +8673,12 @@ async function onConnectButtonClick(e) {
     const config = apiSourceConfig[oai_settings.chat_completion_source];
     if (config) {
         const apiKey = String($(config.selector).val()).trim();
-        if (apiKey.length) {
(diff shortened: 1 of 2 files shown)


Geechan:

Closing for now.

PR #583, opened 2026-06-21
fix: DOM content bleed over user and assistant messages

platberlitz:
# Summary

Bugs: 
The previous assistant messages bleeds through the DOM in the previous user message, additionally, swiping causes the DOM to go to a blank page for a millisecond, goes back to the previous DOM, before it finally generates and shows the current DOM changes.

Fixes: 
1. Mirrors the reasoning.js pattern; before reuse, resets the cache.
2. Syncs messages on regenerates when swiping.


Geechan:

Closing for now.

PR #584, opened 2026-06-21
feat: show main LLM message before pre-generation intercept

platberlitz:
# Summary
Sometimes it takes ages for the pre-generation intercept to complete because it waits for the main LLM output first before changing it (intended behaviour).
To add more ease of mind psychologically and to increase user control, I added a toggle (on by default) to view the main LLM output before proceeding with pre-generation intercepts. User may just use it as is or proceed with the pre-generation intercept.


Geechan:

Closing for now.

PR #586, opened 2026-06-21
revert: restore codex/conversation-shell-polish branch

Geechan:
Reverts PRs #568, #569, #573, #576, #585 as these were causing huge regression problems. Syncs back to branch codex/conversation-shell-polish, which is confirmed to be stable and working.


Geechan:
@platberlitz Update all your new PRs to sync to this PR. We're reverting the mobile refactor as it was causing way too many problems that would take an incredible amount of time to fix on their own.

platberlitz:
> @platberlitz Update all your new PRs to sync to this PR. We're reverting the mobile refactor as it was causing way too many problems that would take an incredible amount of time to fix on their own.

Oh okay. Will do

Geechan:
Merging now as it is a critical bug fix, with approval from @TheLonelyDevil9.

PR #587, opened 2026-06-21
fix: let in-chat agent trim regex actually work

platberlitz:
# Summary
I only noticed just now Trim Regexes in ICA doesn't work, FML. The stuff I'm trying to trim like CHOICES blocks, which should be trimmed at Min Depth 2, didn't get trimmed in context. Now it is. Yay


Geechan (approved):
Easy fix, safe to merge.

PR #591, opened 2026-06-21
feat: show main LLM message before pre-generation intercept

platberlitz:
# Summary

Sometimes it takes ages for the pre-generation intercept to complete because it waits for the main LLM output first before changing it (intended behaviour).
To add more ease of mind psychologically and to increase user control, I added a toggle (on by default) to view the main LLM output before proceeding with pre-generation intercepts. User may just use it as is or proceed with the pre-generation intercept.

### DIFF
--- a/public/scripts/extensions/in-chat-agents/agent-runner.js
+++ b/public/scripts/extensions/in-chat-agents/agent-runner.js
@@ -319,6 +319,13 @@ export function cancelAgentGeneration() {
     return false;
 }
 
+function skipPostMainInterceptChanges() {
+    agentGenerationCancelRevision++;
+    abortActiveAgentRequests('Post-main intercept changes skipped by user.');
+    clearAllPromptTransformRunningToasts();
+    notifyAgentGenerationStateChanged();
+}
+
 function abortActiveAgentRequests(reason = 'Agent generation cancelled.') {
     const error = reason instanceof Error ? reason : new Error(String(reason));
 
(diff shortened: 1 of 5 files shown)


platberlitz:
Ugh not what I intended. It's just working the same as a post-gen... I'll redo

platberlitz:
Okay, now works as intended.

Toggle on ->
Generating initial message ->
Shows the main output first in a window ->
User decides to see if they will proceed with intercept or not.

Geechan:
Unless I'm missing something, the intended behaviour still doesn't seem to be working as it should.

>Toggle on ->
Generating initial message ->
Shows the main output first in a window ->
User decides to see if they will proceed with intercept or not.

Following this, the main body outputs in the chat window without intercepting, then attaches the pre-gen behaviour afterwards without a user prompt.

platberlitz:
> Unless I'm missing something, the intended behaviour still doesn't seem to be working as it should.
> 
> > Toggle on ->
> > Generating initial message ->
> > Shows the main output first in a window ->
> > User decides to see if they will proceed with intercept or not.
> 
> Following this, the main body outputs in the chat window without intercepting, then attaches the pre-gen behaviour afterwards without a user prompt.

That’s very weird. I tested it multiple times last night and it worked.
I’ll test later again.

platberlitz:
Okay, pushed something. Will test as well

platberlitz:
image

Appears to work on my end :+1:

PR #592, opened 2026-06-21
feat: improve accent profile settings

platberlitz:
## Summary

- make Accent Profiles in Appearance settings a collapsible inline drawer with persisted expanded/collapsed state
- keep Save Current visible while profiles are collapsed
- bump seeded accent profiles to 30 total and migrate existing users via the seed version
- More accent profiles because some of them were samey. Added some esoteric ones for the less creative and the lazy (like me)

## Tests
- npm run lint
-  npm run check:frontend-budgets

### DIFF
--- a/public/css/sillybunny-theme.css
+++ b/public/css/sillybunny-theme.css
@@ -4643,6 +4643,20 @@ input[type='range']::-moz-range-thumb {
     gap: 12px;
 }
 
+.sb-accent-profiles-toggle {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: 10px;
+    flex: 1 1 auto;
+    min-width: 0;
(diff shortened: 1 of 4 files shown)


platberlitz:
@Geechan @TheLonelyDevil9 

It works. It was Kimi K2.7 that fixed it of all things, and I spent so much money on frontier to fix this and all it did was overthink and do nothing.

I'm a Kimihead now.

Geechan:
Collapsible drawer is open by default. Ideally, should it be closed by default, to reduce clutter? It's still fairly obvious that the end user can open it later if they need to.

The list is quite long and takes up a lot of real estate. The PR looks good otherwise.

platberlitz:
Should be closed by default now

Geechan (approved):

PR #593, opened 2026-06-21
fix(nav): hide Workspace/Customize toggle in vertical nav layouts

platberlitz:
This PR hides the 'Show Workspace and Customize' toggle when the navigation layout is set to vertical, since those shortcuts are hard-coded into the side rail in that mode.


platberlitz:
Nevermind

PR #598, opened 2026-06-21
fix: improve mobile UI padding, corners, and opacity

platberlitz:
# Summary

**Mobile only**
- Reduced message and chat edge padding
- Softer rounded corners
- Slight paper texture (toggleable, off by default in Mobile dropdown settings)
- Lower top bar height
- Edge-to-edge bottom chat bar and message box
- Theme-aware thought box with less margin


Geechan:
Breaks desktop padding. Thoroughly test this PR on desktop before finalising this PR:

image

Should look like this (current staging):

image

platberlitz:
I swear no matter how many times I say it's mobile-only it touches desktop anyway.

platberlitz:
Should be decoupled on desktop now, I checked. You can check too just in case I missed something (make sure to hard refresh)

PR #599, opened 2026-06-22
fix: make Icons Only work on horizontal side rail

platberlitz:
# Summary
When toggling on the "Icons only in shell tabs", it only applies to the vertical side rail instead of also the horizontal top bar. This fixes that.


Geechan (approved):
Simple line change. Safe to merge.

PR #601, opened 2026-06-22
fix: make regex HTMLs work universally on ICA

platberlitz:
# Summary
- The regexes from the User Stat Agent and the Level Up Companion aren't rendering properly as actual HTML when they're on the Companions Panel. The same regexes work on ICA inline agents when rendered on the chat itself.


platberlitz:
Tested and fixed

PR #602, opened 2026-06-22
fix: prevent switching branches from hanging

platberlitz:
# Summary
Often when I try to switch branches from staging it freezes the screen and I have to close the entire browser and switch from a fresh session after logging in again. Using iOS.

### DIFF
--- a/src/endpoints/server-admin.js
+++ b/src/endpoints/server-admin.js
@@ -1012,10 +1012,19 @@ router.post('/switch-branch', requireAdminMiddleware, async (request, response) 
             await git.stash(['push', '-u', '-m', `Auto-stash before switching to ${branch}`]);
         }
 
-        await git.fetch(['--all', '--prune']);
-        const branchSummary = await git.branch(['-r']);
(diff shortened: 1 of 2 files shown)


src/endpoints/server-admin.js
```diff
@@ -1012,10 +1012,19 @@ router.post('/switch-branch', requireAdminMiddleware, async (request, response)
             await git.stash(['push', '-u', '-m', `Auto-stash before switching to ${branch}`]);
         }
 
-        await git.fetch(['--all', '--prune']);
-        const branchSummary = await git.branch(['-r']);
-        const remoteBranches = getRemoteBranchesFromSummary(branchSummary);
-        const remoteBranch = resolveRemoteBranchName(remoteBranches, branch);
+        // The branch dropdown already refreshed remote refs; resolve locally first
```
Geechan: Remove this LLM comment; not necessary to include.

Geechan (changes requested):

src/endpoints/server-admin.js
platberlitz: Now removed

PR #607, opened 2026-06-22
fix: immediate cache clear for both tokeniser and macros

platberlitz:
# Summary
- Upstream SillyTavern has the issue where you need to refresh, wait for a new message, or toggle and untoggle to clear the tokeniser cache and see the updated values.
- Macros also have the same problem; they stay in the cache if you disable them.
- This PR will automatically clear these caches immediately.

## What Did It Do
- When someone changes the tokenizer or token padding, the app now immediately throws away the old token-count data
- If message token counts are visible, the app recalculates and refreshes them right away.
- If token counts are disabled, it clears the displayed counters instead of leaving old ones around.
- When an extension is disabled, any macros that extension registered are now removed immediately.
- This cleanup works for both the old macro system and the newer experimental macro engine.


TheLonelyDevil9:
Excellent. Lemme just run some checks

platberlitz:
> Excellent. Lemme just run some checks

Sec, still not working on some macros

platberlitz:
Okay it works.
@TheLonelyDevil9 @Geechan 

Go ahead and test - you may use my preset or Gee's
Essentially, it should make it so that if you turn off a {{setvar::}} toggle, it actually turns off and is no longer part of the prompt/context instead of being saved into the cache.

So for Gee's it would be turning on Custom Instructions -> Turning it off without turning on any other macro and see if the main prompt still caches it or not. It shouldn't anymore.

For mine my testing is turn on NSFW Mode -> is now in Main Prompt -> turn it off -> No longer in Main Prompt

PR #612, opened 2026-06-23
fix: SillyBunny generation cancels breaks SB

platberlitz:
# Summary
- Users report: "Repeatedly starting a generation, then stopping it, then starting again will cause SillyBunny to just straight up die" and "sometimes when a response is streaming and I cancel it, it will some break SB and spam the log with error: Client disconnected Client disconnected then I have to start it again but also when a response is streaming like normal and it chokes up half way through streaming the dailogue, it may cause the error to appear too. Error only seems to happens when the streaming generation doesn't finish properly "
- Seems to happen on Windows, but may happen to other OSes as well.

## What Did Gippity Do
https://github.com/platberlitz/SillyBunny/pull/612#issuecomment-4783262303

### DIFF
--- a/src/endpoints/backends/chat-completions.js
+++ b/src/endpoints/backends/chat-completions.js
@@ -36,7 +36,7 @@ import {
     flattenSchema,
     summarizeLlmPayloadForLog,
 } from '../../util.js';
-import { isBenignStreamAbort } from '../../stream-disconnect-guard.js';
+import { isRequestCancellationError } from '../../request-cancellation.js';
 import {
     convertClaudeMessages,
     convertGooglePrompt,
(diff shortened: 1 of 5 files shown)


Geechan (approved):

Geechan:
Latest commit confirmed to fix crashing as reported by the user.

TheLonelyDevil9:
Reviewing, but looks good. Just some checks and will merge

TheLonelyDevil9:
Checking
image

Geechan:
@TheLonelyDevil9 Revert the changes you made. They go out of scope of my initial intent and add unnecessary complexity.

Commit 3970b918d9286813ca1af97b1ecf1b6eeb39e46f does not apply the same logic as the initial PR description, as it proved to be ineffective. It should remain untouched. This is the current description of what it does:

- Ignore benign generation-disconnect crashes at the server process boundary instead of exiting on expected client aborts.
- Add a tiny shared guard for stream-disconnect / request-cancellation errors, with a focused unit test.
- Keep fatal exceptions and real rejections crashing normally, so non-disconnect bugs still surface.

TheLonelyDevil9:
Alright, reverting it

PR #613, opened 2026-06-23
fix: iOS keyboard raising window height

platberlitz:
# Summary

On iOS, the keyboard always raises the window, treating it as though it were a desktop, making the bottom half inaccessible until closing the Customize and Workspace windows again. However, when trying to type on a text input, the keyboard just raises the window again.


platberlitz:
Tested it. It seems to be an iOS problem where it truncates the bottom regardless if the text input is near the bottom of the screen but otherwise it doesn't always truncate it now as long as the text input box is in the middle or above.

Geechan (approved):

PR #616, opened 2026-06-23
docs: update alt language README files

Geechan:
Changes all alternative language README files to actually be about SillyBunny instead of following upstream SillyTavern docs, while letting the main README.md file point to these alt language options.

These alt language files are kept shorter than the main README.md file for now since they're basic LLM translations. They mainly serve to remove irrelevant content that mentions the project being SillyTavern.

No obligation to keep them updated, but should be do-able with LLM translations everytime the README gets updated.


platberlitz (approved):
I don’t know the other languages but you probably did the Japanese one huh? Lol

PR #618, opened 2026-06-26
fix: viewport escaping out of bounds due to dragging in general

TheLonelyDevil9:
This PR fixes the behavior where, when dragging between the QR/GG bar and chat input bar on mobile specifically, the viewport can be fully dragged to the top of the screen (where the keyboard would usually sit when triggered), as well as to the right, where the companion bar would pop out from (it was a large gutter on the bottom and the right side basically)

Tested as fixed by hardening the dragged touches on QR/GG/Composer gap/side edge etc on mobile (firefox latest, Android 16)


platberlitz (approved):
No desktop regression and fix works.

PR #619, opened 2026-06-26
fix(ui): make edit message icons smaller on mobile

platberlitz:
# Summary

Makes the mobile icons smaller for Check, Copy, the bulb icon, trash icon, up, down when editing a message so they don't overflow.


TheLonelyDevil9:
Checking

PR #620, opened 2026-06-26
feat(ui): preset settings dropdown

platberlitz:
# Summary

Creates a dropdown in Presets page for "Keep API/model linked to preset", "Keep sampling settings linked to preset", and "Remember sampling settings per model" a dropdown called "Preset API/Sampler Linking" which is closed by default, the same style as the other dropdowns.


TheLonelyDevil9:
Checking

PR #621, opened 2026-06-27
fix(connection_profile): persist custom endpoint profiles secret

platberlitz:
# Summary

Issue:
"Custom endpoint profiles bugged
Adding keys doesn't link secrets, it links a new credential created when saving the profile instead of what exists
Model picker goes blank, fetch URL error "
This can be mitigated by using Authorization: Bearer (key) in Additional Parameters, but this is a hacky, strange solution.

This PR ensures keys aren't duplicated and whichever secret is used for the custom endpoint profile is actually saved and attached to the profile.

### DIFF
--- a/public/scripts/extensions/connection-manager/index.js
+++ b/public/scripts/extensions/connection-manager/index.js
@@ -15,7 +15,7 @@ import { SlashCommandScope } from '../../slash-commands/SlashCommandScope.js';
 import { cancelDebounce, collapseSpaces, getUniqueName, isFalseBoolean, isTrueBoolean, uuidv4, waitUntilCondition } from '../../utils.js';
 import { t } from '../../i18n.js';
 import { getSecretLabelById } from '../../secrets.js';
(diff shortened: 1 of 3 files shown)


TheLonelyDevil9:
Oh, thanks for this one, reviewing

TheLonelyDevil9:
As per Gippity:

```
Blocking: public/scripts/extensions/connection-manager/index.js:624 only saves the selected custom endpoint preset’s secretId, but applying a saved profile still never reselects/binds that custom endpoint preset. openai.js:6955 and openai.js:5399 still prefer selected_custom_endpoint_preset.secretId, so status/model fetches and normal generation can use the wrong secret if another custom endpoint preset is selected. This leaves the stated “model picker goes blank, fetch URL error” path only partially fixed.
Standards: public/scripts/extensions/connection-manager/index.js:593 adds fork-specific behavior in a base file without the required short SillyBunny divergence comment from AGENTS.md/CONTRIBUTING.md.
```

Fixing unless that sounds wrong

PR #624, opened 2026-06-27
feat(qol): add a Delete + Add Swipe button

platberlitz:
# Summary

Adds a button that "Deletes this swipe + adds a new swipe" to the right of the pencil icon and to the left of the trash icon in a message.

Reasoning: Sometimes a swipe fails and it's kind of messy to delete this swipe and swipe again. This is just a button that does both easily.


TheLonelyDevil9:
Been doing it manually for so long, good change. Let me run some checks and merge after.

PR #626, opened 2026-06-27
fix: Hardening viewport escaping across the board

TheLonelyDevil9:
Will likely need extensive testing to ensure no unintended breaks, but basically I've tried to harden the viewport escapes that keep happening across mobile and PC both. This sets up a viewport anchor so accidental page-level scrolls always snap back to 0. Additionally, it changes the direct scrollIntoView() function with scoped scrolling functions instead.

Finally, all scanned, non-vendored plain .focus() calls are now converted to focus ({preventSCroll: true}) functions. This should HOPEFULLY cause less of these viewport drifts that put the top menu out of reach, and show you black spaces with nothing in it.

### DIFF
--- a/public/script.js
+++ b/public/script.js
@@ -11893,7 +11893,7 @@ async function promptForMessageScreenshotRange(messageId) {
         onOpen: function () {
             const startInput = popup.dlg.querySelector(`#${MESSAGE_SCREENSHOT_INPUT_IDS.start}`);
             if (startInput instanceof HTMLInputElement) {
-                startInput.focus();
+                startInput.focus({ preventScroll: true });
                 startInput.select();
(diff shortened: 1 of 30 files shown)


platberlitz:
Fixed some regressions.

- Added missing `getStreamingReasoningRenderInterval` export.
- Updated `shouldReduceStreamingDomWork` to honor `iosEnabled` and `androidEnabled`.
- Added missing Android power-user defaults.
- Restored real scroll fallback in In-Chat Agents.
- Added focused tests for the regression paths.

platberlitz:
Tested on both desktop, iOS, Android. No glaring regressions and haven't seen much viewport escapes.

PR #629, opened 2026-06-27
feat(tts) - add stop button for TTS playback

kilonova1832:
Small PR to add a stop button to TTS playback
What it does -
- When playback starts, the narrate icon is replaced by a stop icon
 - if playback ends normally the icon returns to the normal narrate icon automatically
 - playback can be stopped before it ends by clicking the stop icon. This should stop TTS playback immediately and return the icon back to it's original state.
 
 Without this there's no way to stop TTS playback once it has started.
 
 (Disclaimer - code is AI assisted but this overview is not)


platberlitz (approved):
Very simple and functional, I would say this is good to merge.

PR #630, opened 2026-06-28
fix(ica): render macros in Companion Agents drawer

platberlitz:
# Summary

## Issue
In the draggable Companion Agents drawer, macros such as {{user}}, {{char}}, etc are rendered as plaintext rather than converted through the DOM.


platberlitz:
Tested and works.

PR #632, opened 2026-06-28
feat(conversation): clickable avatar to show full picture

platberlitz:
You should be able to click character avatars in Conversation Mode to show the full picture (pops out full picture, X button to close), same as other parts of SB.


TheLonelyDevil9:
Changed a few things GPT picked up:
- Header avatars also open the full picture now.
- The Pals rail still behaves normally: clicking a row opens the chat instead of accidentally opening the picture.

PR #633, opened 2026-06-28
fix(sampling_profile): automatic sampling profile switch

platberlitz:
# Summary

## Issue
When you switch to a different model using the backend dropdown (e.g. going from ZAI to MoonshotAI), it doesn't automatically load the sampling profile associated with the model. You have to switch to a different model, then back to the model you want to use to load the sampling profile.

## What This Changes
It should instantly switch to the sampling profile without having to switch models and load it manually.


platberlitz:
Tested and works.

PR #635, opened 2026-06-28
fix(custom_css): allow importing of custom CSS when uploading themes that have it

platberlitz:
# Summary

## Issue
A regression that made it so even when you say yes to importing a custom CSS bundled in a theme json, nothing happens.

## What This Fixes
Reverts the regression.


platberlitz:
Tested and works.

PR #636, opened 2026-06-28
fix(mobile_webkit_safari): steady the input textbox viewport

platberlitz:
# Summary

## Issue
Specifically on *Safari*, and it doesn't happen on Firefox iOS, when I try to type on an input box, it steadies the screen and doesn't move to the input box, causing it to go down whenever I type a letter, then up, then down again, etc.

## What This Fixes
Stabilises input box viewport


platberlitz:
Tested and works.

TheLonelyDevil9:
Reviewed. Merging

PR #637, opened 2026-06-29
feat(ica): include the number of tokens used by agents in the UI

platberlitz:
# Summary

## Issue
While you can see token counts of separate agents by clicking the Main prompt, there isn't really a proper, easy UI to see how much tokens agents use in total.

## What This Adds
Multiple ways to access In-Chat Agents token counts.
- In Prompt Manager
- In agents page

## For Companion Agents?
Adds Input: xx Output: xx token counts on the side drawer.

### DIFF
--- a/public/scripts/PromptManager.js
+++ b/public/scripts/PromptManager.js
@@ -39,6 +39,8 @@ function debouncePromise(func, delay) {
 const DEFAULT_DEPTH = 4;
 const DEFAULT_ORDER = 100;
 const PROMPT_MANAGER_DESKTOP_SPLIT_QUERY = '(min-width: 961px)';
+// SillyBunny In-Chat Agents reserve this extension prompt key prefix.
+const IN_CHAT_AGENT_PROMPT_KEY_PREFIX = 'inchat_agent_';
 
 /**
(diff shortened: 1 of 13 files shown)


platberlitz:
image
image
image

TheLonelyDevil9:
As per GPT:

[P2] Batched companion input tokens are stored as per-agent usage: [companion-runner.js#L1271-L1274](https://github.com/platberlitz/SillyBunny/blob/3c39fd2fab0c412223ce506e78138d1019d1a594/public/scripts/extensions/in-chat-agents/companion/companion-runner.js#L1271-L1274). batchInputTokens is the full shared batch request, but every companion result gets that same value and the UI renders it as that companion’s Input count. For a two-companion batch, the drawer/dashboard makes the shared input look doubled if users compare rows. 


```
Either label it as shared batch input, store it once at batch level, or estimate/split per companion. The added test only covers a single companion run, so this path needs a batch-token regression test.
```

Didn't want to be heavy-handed with your contribution, let me know if there's a direction you want here

platberlitz:
Split per companion is the intention.

PR #638, opened 2026-07-01
feat(conversation_mode): decouple conversation mode from other features using Rest API

platberlitz:
# Summary
Before, Conversation Mode worked using the following paths:

What works today:
POST /api/settings/get and POST /api/settings/save can read/write the saved Conversation Mode store under extension_settings.sillybunny_conversation.
POST /api/groups/create can create group Conversations with conversation_settings.
POST /api/backends/chat-completions/generate can generate text if you supply the full model payload yourself. 

The issue is that makes it too tied with group conversations, instead of being decoupled and considered its own thing. This will allow JSON-only curl commands, when the Rest API is added, so that they can be called without touching other parts of SillyBunny.

## What Endpoints Are added?
- POST /api/sillybunny-conversation/store/get
- POST /api/sillybunny-conversation/store/save
- POST /api/sillybunny-conversation/thread/get
- POST /api/sillybunny-conversation/thread/save
- POST /api/sillybunny-conversation/message/append
- POST /api/sillybunny-conversation/message/send
- POST /api/sillybunny-conversation/info

## What Else?
- Still the same Conversation Mode settings, router is in `src/server-startup.js`. Conversation Mode chat/textbackend generation handlers were refactored as well, ensuring that duplication between group chats and Conversation Mode doesn't happen.
- Adds per-persona storage.
- Adds TTS support.

### Bug Fixes
- Made replying actually show up properly. 
- Screenshot example: image
- Makes group DMs show up even without sending a message first.
- Mitigates/removes character impersonation in group chats.
- Uses system time on your computer/phone to inform what time it is to the characters.
- Fixed manual selfie generation.

## What To Test?
Essentially, just making sure that it still functions exactly the same and isn't tied to group chats. That means when accessing a Conversation Mode chat from the Recent Chats in the landing page, it should go directly to the Conversation Mode chat, not a roleplay-style chat that just has a skin on it.


TheLonelyDevil9:
Been using this on my personal fork for a week, no regressions noticed

platberlitz:
## What was fixed in the latest commits
**Security and server-side safety:**
- Closed a hole where the server could be tricked into fetching images from private internal network addresses.
- Fixed rate limiting so one user hammering the server can't block *other* users, and can't dodge their own limit by changing IP address.
- Made sure old saved chat data (from before this feature) still loads instead of being rejected as "invalid", no one loses their chat history.

**"Wrong person" bugs:** A whole bunch of bugs where, if you switched personas or characters *while* something was still loading, the result landed in the wrong place. For example:
- A reply meant for character A could get saved into character B's chat.
- Closing the settings panel could overwrite another character's settings.
- A generated selfie could use the wrong character's appearance settings. All of these now remember *who* they were for when they started, and finish for that person.

**Messages silently disappearing:** Several situations where your message got eaten with no reply and no error - e.g. sending a DM and then doing something else within 5 seconds, or pinning a message you just sent. These now retry properly instead of giving up.

**Group chats:** Old group conversations could show up empty, hidden, or open as one-on-one chats. Replying to a specific person in a group now actually makes *that* person answer, not whoever spoke last.

**Voice (text-to-speech) and image generation:** These extensions were being loaded a second time behind the extension manager's back, causing doubled audio, duplicate behavior, and running even when you'd disabled them. They're now accessed through one shared, properly-managed channel. Also: background chat audios no longer talk over your current one, and audio starts faster instead of waiting too long.

**Small stuff:** Buttons on mobile were too small to tap reliably (now 44px), a hidden "Companion" menu item was still clickable, unread-message counts now show correctly per chat branch.

Geechan (approved):

PR #642, opened 2026-07-04
fix(ipad_webkit_viewport): steady text box input viewport on iPadOS

platberlitz:
# Summary
## Issue
Previously, the PR https://github.com/platberlitz/SillyBunny/pull/636 was able to fix the viewport constantly jittering up whenever the keyboard appears and hides the input textbox. However, it seems this doesn't work on iPadOS, only on iOS.

## Fix
Applies the same fix to iPadOS.

### DIFF
--- a/public/css/sillybunny-mobile-shell.css
+++ b/public/css/sillybunny-mobile-shell.css
@@ -897,6 +897,20 @@
     }
 }
 
+/* SillyBunny: mirror the <=768px drawer keyboard inset on wide iOS viewports
+   (iPadOS desktop-mode Safari reports a desktop UA and exceeds 768px).
+   .sb-ios-keyboard-inset-active is only toggled while the iOS software
+   keyboard is open, so non-iOS and keyboard-closed layouts never match. */
+html.sb-ios-keyboard-inset-active #left-nav-panel.openDrawer .sb-shell-panel-scroller,
+html.sb-ios-keyboard-inset-active #user-settings-block.openDrawer .sb-shell-panel-scroller,
+html.sb-ios-keyboard-inset-active .sb-shell-root.openDrawer .sb-shell-panel-scroller,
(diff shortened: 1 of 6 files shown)


platberlitz:
Confirmed to have fixed the bug. Safe to merge.

TheLonelyDevil9 (approved):
Reviewing and fixing some stuff, then merging

@Geechan Webkit related and minor enough changes so approving here

Per Fable:

Pushed two follow-ups to this branch:

**Wide-viewport keyboard inset (6fb46a2)**

`--sb-ios-keyboard-bottom-inset` was only consumed inside `@media (max-width: 768px)` blocks, so drawer and panel scrollers on iPads wider than 768px (all landscape iPads, larger iPads in portrait) never got the extra scroll room while the keyboard is open. `syncIOSKeyboardBottomInset()` now toggles `sb-ios-keyboard-inset-active` on the root element while the inset is non-zero, and a width-independent rule mirrors the existing scroller padding for that state. Non-iOS browsers never receive the class, and the padding reverts when the keyboard closes.

**About the test-only commit (e8442ad)**

`openai-proxy-preset-wiring.test.js` was failing on staging: the custom endpoint save flow already trims the key input via `keyInputValue`, but the assertion still expected the old `key: $('#api_key_custom').val()` string. The commit realigns the assertion; no runtime behavior changes. Since it's unrelated to the iPad fix, suggest adding a line to the PR body:

> Also realigns the custom endpoint key assertion in `openai-proxy-preset-wiring.test.js`, which was left stale by the key-trim change.

**Testing**

- Full Jest unit suite: 155 suites / 1444 tests pass
- ESLint and frontend budget checks clean
- The gate removal only widens behavior: `isIOSWebKitPlatform()` returns false on desktop macOS Safari (`maxTouchPoints` is 0), so nothing changes outside iOS/iPadOS WebKit

PR #644, opened 2026-07-08
fix(ica): minimum context overrides number of messages in Companion Agents

platberlitz:
# Summary
## Issue
Seen most easily in the Memory Shard agent: instead of adhering to the minimum context before it runs, it runs based on that *but* only bases the summary on the number of messages it's set to (by default, 30 messages), making it miss additional context from beginning to end, especially if there are more than 30 messages before it hits 30k context.
Additionally, it considers hidden messages in summarising and context as well when it shouldn't.


TheLonelyDevil9:
Good to go. Merging.

PR #647, opened 2026-07-08
fix: Gemini thought signatures getting mangled/corrupted

TheLonelyDevil9:
What was happening earlier: Gemini thought signatures were getting corrupted for unknown reasons on both API as well as reverse proxy modes. This is now fixed with some targeted changes.

In Claude's terms for reference:

- **Bug:** Gemini 2.5/3 models (e.g. 3.1 Pro Preview) intermittently rejected requests with `400: Corrupted thought signature`, both direct and through reverse proxies (the payload SillyBunny builds is the problem; proxies just relay it). 
Three defects in `convertGooglePrompt`:
  1. The stored text signature was stamped onto **every** text part of a message, and onto **user-role** messages that carried a stray signature — signatures are only valid once, on model turns.
  2. The consecutive-same-role merge double-handled signed text parts: the text was concatenated into the previous part **and** the part was pushed again — duplicated content with misplaced signatures. Triggers in group chats and whenever assistant-role prompt injections sit next to a signed chat message.
  3. Plain text could be folded into a signed part (and signed text into plain parts), mutating content the signature was bound to.
- **Fix:** The signature is attached only to the last text part of model-role messages; the merge keeps signed parts byte-identical as standalone parts and only concatenates unsigned plain text. Gemini 3 `skip_thought_signature_validator` fallbacks for function calls and image models are unchanged.
- **Verification:** A scenario harness against the old code reproduced all three defects; the fixed code passes all 11 checks, and the existing 152 prompt-converter tests still pass.

### DIFF
--- a/src/prompt-converters.js
+++ b/src/prompt-converters.js
@@ -584,10 +584,15 @@ export function convertGooglePrompt(messages, model, useSysPrompt, names) {
         // Inject stored thought signatures, or fall back to bypass magic for Gemini 3
         if (/gemini-3/.test(model) || /gemini-2\.5/.test(model)) {
             const skipSignatureMagic = 'skip_thought_signature_validator';
-            const textSignature = message.signature;
+            // Signatures are only valid on model turns; stamping one on a user-role part
+            // gets the request rejected with a 400 "Corrupted thought signature".
+            const textSignature = message.role === 'model' ? message.signature : null;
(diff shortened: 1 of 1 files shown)


TheLonelyDevil9:
As this is a bugfix that's critical, skipping the review @Geechan @platberlitz

PR #649, opened 2026-07-08
feat: add LinkAPI chat completion provider integration with auto-routing

TheLonelyDevil9:
I've added LinkAPI as a chat completion provider in SB.
One provider entry covers the gateway's three API formats.

---
As per Claude:
Requests are routed automatically by model id — claude-* uses the Anthropic-native /v1/messages relay, gemini-* uses the Gemini-native /v1beta/...:generateContent relay, and everything else (gpt/deepseek/grok/glm, plus bracket-tagged flat-rate per-request channels like [SP].../[次]...) uses the OpenAI-compatible /v1/chat/completions relay.
Includes a region selector (Global linkapi.ai, US api.linkapi.ai, HK hk.linkapi.ai), a dynamic model list, and standard secret storage.
---

For the why: Sick of juggling custom profiles? This provider has you sorted based on just the key. Switch keys > Switch groups basically.

---
Per Claude:
LinkAPI exposes native Anthropic/Gemini endpoints alongside the OpenAI-compatible one, and its docs recommend the native formats for reliable tool calls, streaming, and thinking output. Using it through the generic Custom endpoint loses all of that (plus Claude prompt caching, prefill, thinking budgets, and Gemini thought signatures).
A dedicated source with per-model format routing gets full native behavior without asking users to juggle three Custom profiles. 
LinkAPI keys are group-scoped upstream (a key's group decides pricing/channels and which models `/v1/models` returns), so the existing multi-key secret storage covers "one labeled key per group" with no new UI.

### Technical deets:
- **New shared module `public/scripts/linkapi-utils.js`** (imported by both client and server, same pattern as `local-url-utils.js`): region → base-URL map plus `getLinkApiRequestFormat(model)`. Bracket-tagged ids short-circuit to `openai` because those per-request channels only relay through the OpenAI-compatible endpoint.
- **Server (`chat-completions.js`)**: the generate switch delegates claude/gemini models to the existing `sendClaudeRequest` / `sendMakerSuiteRequest` by synthesizing their reverse-proxy override fields (`reverse_proxy` = LinkAPI base, `proxy_password` = LinkAPI key) — no duplication of the native handlers, so Claude caching/thinking/websearch and Gemini streaming work unchanged. OpenAI-format models fall through to the shared `chat/completions` path; `/status` reuses the shared Bearer `/models` fetch.
- **Client**: streaming (`getStreamingReply`), reasoning extraction, and structured-output extraction detect the payload shape (`candidates` → Gemini, `choices` → OpenAI, else Anthropic deltas) rather than assuming one format, since one source can return any of the three. Generation params send the Claude-only / Gemini-only fields (sysprompt, prefill, top_k, stop-string limits) only for the matching leg. Context caps follow the underlying model family (claude → 200k, gemini → 1M, else 128k) independent of wire format, with the unlocked-context checkbox enabled for this source; the temperature cap is 1.0 on native legs and 2.0 on the OpenAI leg.
- Everything else follows the established Z.AI/SiliconFlow sub-endpoint provider pattern (enum entries, `settingsToUpdate`, model-id search control, connect config, tokenizer/tool-calling/slash-command/auto-connect registrations).

### Testing?

- `node --check` and ESLint pass on all 14 touched files on this branch; `getLinkApiRequestFormat` probed directly (`[SP]claude-opus-4-7 → openai`, `claude-sonnet-4-5 → anthropic`, `gemini-3.5-flash → google`, `gpt-5.5 → openai`).
- The identical change set was runtime-tested on the development tree it was authored on: server boots clean; browser check confirmed the source appears in the dropdown, the form shows/hides correctly, and settings persist with no new console errors.
- Keyless probes against the live backend confirmed each routing leg reaches the intended code path (`/status` → shared models fetch; claude model → native handler key check; gpt model → shared OpenAI-compatible chain).
- Not yet run: end-to-end generation against LinkAPI with a real key (all three legs, streaming, tool calls). The native handlers themselves are untouched, so residual risk is concentrated in the new routing/param plumbing.
---

Let me know if screenshots are needed, though nothing new on the UI is epected.
The UI addition is a standard provider settings form (key input, region dropdown, model picker) built from the same markup as the existing Z.AI form, with no visual changes elsewhere.

### DIFF
--- a/.github/readme.md
+++ b/.github/readme.md
@@ -5,6 +5,12 @@
 <img src="screenshots/banner.jpg" width="100%">
 </div>
 
+<div align="center">
+
+English | [Deutsch](.github/readme-de_de.md) | [中文](.github/readme-zh_cn.md) | [繁體中文](.github/readme-zh_tw.md) | [日本語](.github/readme-ja_jp.md) | [Русский](.github/readme-ru_ru.md) | [한국어](.github/readme-ko_kr.md)
+
+</div>
+
 ---
 
 An elegant fork of [SillyTavern](https://github.com/SillyTavern/SillyTavern), designed with a cleaner, graphical shell UI; Bun-based backend; built-in tutorials, presets, extensions, and a quick-start dashboard; and a lightweight agentic system to facilitate modern agent functionality.
--- a/public/index.html
+++ b/public/index.html
@@ -342,7 +342,7 @@
                             -->
                             <div id="range_block_novel">
                                 <div class="flex-container gap10h5v justifyCenter">
-                                    <div data-source="openai,claude,aimlapi,openrouter,ai21,makersuite,vertexai,mistralai,custom,cohere,perplexity,groq,siliconflow,minimax,electronhub,chutes,nanogpt,deepseek,xai,pollinations,moonshot,fireworks,cometapi,azure_openai,zai,workers_ai" data-tg-samplers="temp" class="alignitemscenter flex-container flexFlowColumn flexBasis30p flexGrow flexShrink gap0">
+                                    <div data-source="openai,claude,aimlapi,openrouter,ai21,makersuite,vertexai,mistralai,custom,cohere,perplexity,groq,siliconflow,minimax,electronhub,chutes,nanogpt,deepseek,xai,pollinations,moonshot,fireworks,cometapi,azure_openai,zai,workers_ai,linkapi" data-tg-samplers="temp" class="alignitemscenter flex-container flexFlowColumn flexBasis30p flexGrow flexShrink gap0">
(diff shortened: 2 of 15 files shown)


TheLonelyDevil9:
Note on the second commit: `check-readme-mirror` currently fails on bare `staging` and `main` — `.github/readme.md` is missing the language-selector block that was added to `README.md`, so every PR inherits the red X. Included the mechanical 6-line regeneration (`bash scripts/sync-readme-mirror.sh`) here to unblock CI; happy to drop it if you'd rather fix the mirror separately.

platberlitz (approved):

PR #650, opened 2026-07-09
fix(ica): viewing the in-chat agents injected into the preset should show their actual names

platberlitz:
# Summary
## Issue
In-Chat Agents show the filenames whenever you preview them in the preset (e.g. inchat_agent (bunch of numbers)) instead of the actual title.

## Fix
Should show the *actual* title of the agent, such as "Event Tracker", "Plot Compass", rather than the file names.


platberlitz:
image

Works as seen here.

PR #651, opened 2026-07-09
fix(card_imports): fixes chub AI import API + botbooru support

platberlitz:
# Summary
## Issue
Trying to import Chub links gives the error "Custom import error".
Reasoning was that Chub for some reason sends webp instead of regular PNG files.

## Fix
Fixes the API so it works correctly.

## Additions
Also adds Botbooru support.


platberlitz:
image

Finally it fucking works.

PR #654, opened 2026-07-09
feat: New ordering mechanism for agents and companion sidebar

TheLonelyDevil9:
As requested, baked in the ordering for companion agents and the agents UI itself with a drag handle. Summary to follow.

Per Claude:
- **Agents list**: the per-card grip handle is now always visible (it was touch-only) and drag-anywhere still works on desktop. Focusing the grip and pressing Arrow Up/Down nudges a card one slot for precise placement without dragging.
- **Companion panel**: every companion section gets the same grip in its header; the panel body is drag-sortable with the same arrow-key nudging.
- **Single source of truth**: both surfaces persist through a new agent-store helper, `reorderAgentsIntoOrderSlots()`. It deals the dragged subset back into the order-sorted slots it already occupies and renumbers every agent to `index * 10`, so reorders always produce clean, collision-free `injection.order` values (pre-existing duplicate orders resolve on the first reorder). The agents page's `reorderAgentsInGroup()` now delegates to it.
- **Cross-surface sync**: reordering in the panel refreshes the agents list via a new `refreshAgentList` hook, and reordering in the agents list refreshes the open panel via `refreshCompanionPanel()`.
- Visual order and sequential execution order remain the same field, as before.

Also included: swiping into a brand-new swipe slot no longer shows the previous swipe's companion state. New slots have no `swipe_info` entry yet, so companion reads fell back to `message.extra` and displayed stale results until the companions overwrote them; the carried-over results are now cleared on `MESSAGE_SWIPED` so message cards and the panel go blank while the new reply generates. Existing-swipe navigation still restores each swipe's own stored results.

# Summary (Added by Pura)
- Reordering Companion Agents is now easier through the draggable area in Companion Agents drawer.
- Exposes Order number for Companion Agents in the UI settings.

### DIFF
--- a/public/scripts/extensions/in-chat-agents/agent-store.js
+++ b/public/scripts/extensions/in-chat-agents/agent-store.js
@@ -1618,6 +1618,59 @@ export async function saveAgent(agent) {
 }
 
 /**
+ * Applies a new visual order to a subset of agents. Subset members are dealt back into the
+ * order-sorted slots the subset already occupies (agents outside the subset keep their relative
+ * position), then every slot is renumbered to `index * 10` so drag-and-drop always produces
+ * clean, collision-free order values.
+ * @param {string[]} orderedSubsetIds Agent ids in their new visual order.
+ * @returns {Promise<boolean>} Whether any agent's order value changed.
+ */
+export async function reorderAgentsIntoOrderSlots(orderedSubsetIds) {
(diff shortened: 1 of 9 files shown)


platberlitz:
Unable to drag and drop in the Companions side drawer. Fixing.

PR #658, opened 2026-07-10
fix(mobile): hold the chat composer above the iOS keyboard to stop the viewport escape

TheLonelyDevil9:
On iPhones, focusing the chat textbox no longer lets Safari pan/scroll the app off screen.
While the composer is focused, the mobile shell now gives up the keyboard's height — keeping its stable layout top — so the composer sits directly above the keyboard and the top bar stays visible while typing.

Per Claude, other deets:

## How?

- New `getComposerKeyboardInset()` in `sillybunny-tabs.js`: while the composer is focused on an iOS mobile viewport, `getShellViewportSize()` returns the stable layout viewport minus the keyboard height, so the caret is never behind the keyboard and Safari has nothing to reveal.
- The last measured keyboard height pre-shrinks the shell on `focusin`, before the keyboard finishes animating — that is what prevents the caret-reveal. A 700 ms settle window plus a follow-up sync converge on the measured height and absorb mid-animation resize jitter (the jitter that originally motivated moving the composer into stable mode).
- If Safari already panned (first focus of a session, keyboard height unknown yet), the shell refuses to shrink — shrinking while panned reproduces the exact escape geometry — and keeps today's clamped-pan behavior for that single focus.
- While the composer is held clear (root class `sb-ios-composer-keyboard-inset-active`), `browser-fixes.js` re-arms the document scroll anchor mid-edit: a reset can no longer hide the caret, so any Safari force-scroll snaps back immediately instead of sticking until blur.
- `focusin`/`focusout` now trigger the mobile viewport state sync, so stable-viewport decisions cannot linger on stale focus state when the keyboard stays open across focus moves.
- Android and iPadOS desktop-mode paths are untouched; all new logic is gated on `isIOSWebKitPlatform() && isMobileViewport()`.

## Testing?

- Full unit suite passes: 156 suites / 1474 tests, including a new wiring test block covering the inset logic, the already-panned guard, the pre-shift wiring, and the browser-fixes re-arm.
- ESLint clean on all touched files; frontend budgets unaffected.
- **Not yet verified on a physical iPhone** — needs the reporter's retest on staging. Expected behavior: top bar stays on screen with the composer directly above the keyboard while typing.


platberlitz:
Fuck me i hate webkit

platberlitz:
YESSSSSSSSSSSSSSS TERRA FIXED IT!
Merging.

PR #660, opened 2026-07-11
fix: favorites save errors, lit fav state, and scrollable hotswap bar

TheLonelyDevil9:
## What?

Fixes the chain of bugs that made favoriting characters appear broken:

- `TavernCardValidator` now reports the validation error from the spec the card actually declares, instead of whichever spec branch was tried last (a v2 card missing `character_version` used to surface as a meaningless `Field: spec` error from the v3 branch).
- The bulk-edit favorite flow no longer flips the `is_fav` UI state when the `/api/characters/merge-attributes` save fails — previously the list showed a favorite that was never persisted.
- The favorites hotswap bar scrolls sideways with the mouse wheel, its favorite cap is raised from 25 to 250, and the character/group editor FAV buttons now visibly light up under themes that pin `.menu_button` colors with `!important`.

## Why?

Any card with a spec-conformance gap (e.g. missing `character_version`) failed **every** save — including simple favorite toggles, since `merge-attributes` re-validates the whole card — and the masked error (`Field: spec`) pointed users at the wrong cause. On top of that, two UI bugs made the feature look broken even when saves succeeded: the fav icon never lit under themes that pin `.menu_button` colors (e.g. MoonlitEchoes), and favorites past 25 silently never appeared in the hotswap bar.

## How?

- **Validator** (`src/validator/TavernCardValidator.js`): `validate()` captures each branch's error as it tries V1 → V2 → V3, clears the stale error on success, and on total failure picks the error matching `card.spec`.
- **Bulk edit** (`public/scripts/BulkEditOverlay.js`): early `return` after the error toast.
- **Hotswap bar** (`RossAscends-mods.js`, `sillybunny-tabs.css`): native non-passive `wheel` listener translating vertical wheel to horizontal scroll — `scrollTo(..., behavior: 'instant')` because smooth scrolling makes rapid ticks overwrite each other; trackpad horizontal swipes are left to native handling. `overflow-x: auto` with the scrollbar hidden so the 36px row keeps its height. `FAVS_LIMIT` 25 → 250 is safe because the bar scrolls and avatars are lazy-loaded thumbnails.
- **Fav button styling** (`sillybunny-tabs.css`): `!important` on the two-ID `.fav_on` rules, since theme-level `.menu_button { color: ... !important }` otherwise wins the cascade and renders `fav_on` pixel-identical to `fav_off`. Cache-bust param bumped per repo convention.

## Testing?

- Live verification on a running instance: favoriting a previously-failing card persists and the FAV button lights up under MoonlitEchoes; wheel scrolling moves the hotswap bar in both directions and clamps at the ends.
- Validator fix verified by feeding deficient v2/v3 cards: errors now name the real missing field (e.g. `data.character_version`) instead of `spec`.
- 144 card chunks from a personal library re-validated through the patched validator: all pass.
- Not covered: no automated tests were added; Firefox `deltaMode === 1` wheel handling is untested (converted at ~34px/line).

## Screenshots

None attached — changes were verified live in a browser session (fav button lit state, error toast wording, bar scrolling). Can add before/after captures on request.

## Anything Else?

- Root cause of the failing saves was non-conformant cards produced by an external pipeline; those card files were repaired on the user side (not part of this repo). This PR makes SillyBunny diagnose such cards correctly and keeps the UI honest about failed saves.
- `FAVS_LIMIT = 250` is a sanity bound, not a layout constraint anymore.

### DIFF
--- a/public/css/sillybunny-tabs.css
+++ b/public/css/sillybunny-tabs.css
@@ -2190,7 +2190,16 @@ body.sb-shell-resizing * {
     margin: 0;
     align-items: center;
     flex-wrap: nowrap;
-    overflow: hidden;
+    overflow-x: auto;
+    overflow-y: hidden;
+    /* Wheel-driven sideways scroll; a visible scrollbar would eat the 36px row,
+       and smooth behavior makes rapid wheel ticks overwrite each other */
+    scrollbar-width: none;
+    scroll-behavior: auto;
+}
+
+#right-nav-panel #HotSwapWrapper .hotswap::-webkit-scrollbar {
+    display: none;
 }
 
 #right-nav-panel #HotSwapWrapper .hotswap > small {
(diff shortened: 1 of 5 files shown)


TheLonelyDevil9:
I will resubmit, Claude is being very smart with some shit.

PR #661, opened 2026-07-11
fix: Favs not lighting up, scrolling favorites on desktop, validator behavior changes, etc

TheLonelyDevil9:
This PR focuses on ~~three~~ ~~six~~ ~~seven~~ eight (added by Pura) things:

1- Making sure the card spec validator errors correctly on benign actions (current state was that the error was card spec V3's no matter what error was actually occurring)
2-  If a bulk favorite action was performed, even on failed card validation checks (like the above point fixes), the previous behavior would mark `is_fav` anyway, even if it didn't save. This makes the failure clear at least.
3- The hotswap bar actually scrolls now with favorites populating it. If you've got more than it can display, that is. There used to be a 25 fav limit. Now that's 250.
4- Hotswap bar centering and cleanup
5- Editor Avatar preview - image is now always going to show the entire character image instead of just a cropped variant
6- Favorites strip vs. mode toggle/close button - CSS to ensure the favorites bar and close/mode switch don't overlap. Also, the mode toggle pills have the cursor change to a pointer.
7- Fixing scrolling of said favorites on mobile after the fixes above.
8- Centers icons on the character drawer no matter what, so the damn avatars don't go to one edge or another randomly.

Code review done:
- Scrolling over a long Favorites bar no longer disables Ctrl/Command zoom.
- Normal mouse-wheel scrolling still moves the Favorites bar sideways.
- Fixed the two CI failures caused by the CSS change and stylesheet cache refresh.
- Added an automated browser test so this zoom issue stays fixed.

### DIFF
--- a/public/css/sillybunny-mobile-shell.css
+++ b/public/css/sillybunny-mobile-shell.css
@@ -507,6 +507,18 @@
         padding: 3px 10px !important;
     }
 
+    /* The collapsed shell header parks the mode toggle + close button over the
+       right end of this band, and the characters button floats on its left;
+       inset the favorites block so the divider and strip sit cleanly between
+       the flanking controls instead of running underneath them.
+       Left 48px clears the 44px characters button; right 134px clears the
+       82px mode toggle + 36px close button and their edge offsets. */
+    #right-nav-panel.openDrawer #HotSwapWrapper {
+        margin: 0 134px 0 48px !important;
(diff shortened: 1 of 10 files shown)


platberlitz:
Haha you said six seven

TheLonelyDevil9:
Damn it

platberlitz (approved):

PR #662, opened 2026-07-12
fix(chat_files): hardens chat files, fixes backups and chat renaming

platberlitz:
The following issues are addressed by this PR:

1. Using the LLM auto-rename in the bottom chat bar gives a corrupted brackets name instead of an actual name when using a non-frontier model.
2. Renaming, if successful, can cause the entire chat to be deleted and start from the beginning.
3. Checking through the chats with the notebook icon in the bottom chat bar, there is a Backups button. Clicking it does nothing even though the backups folder definitely shows backup files.
4. Ensures that backups are immediately used when chats get corrupted/deleted.

## Fixes
- Fixed AI chat renaming so invalid outputs like `{}` or `[object Object]` are rejected.
- Made chat renaming safer so it no longer rewrites the character or accidentally resets the conversation.
- Added rollback protection if part of a rename fails.
- Fixed the Backups button so it shows loading, errors, and available backup files.
- Added an immediate recovery copy whenever a chat is saved.
- Automatically restores chats that become corrupted or unexpectedly disappear.
- Prevented intentionally deleted chats from being automatically restored.
- Added security checks and extensive tests for renaming, backups, and recovery.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -120,6 +120,8 @@ Status: Unreleased. This section is currently a mechanical staging PR ledger gen
 - PR #657 (2026-07-11) `feat(agents): let post-generation passes target companion outputs, impersonation text, and picked targets`
 - PR #658 (2026-07-11) `fix(mobile): hold the chat composer above the iOS keyboard to stop the viewport escape`
 - PR #661 (2026-07-12) `fix: Favs not lighting up, scrolling favorites on desktop, validator behavior changes, etc`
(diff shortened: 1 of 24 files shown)


TheLonelyDevil9 (approved):
Reviewed the final recovery lifecycle fixes, route-level regressions, retention bounds, and safety behavior; no remaining merge blockers found.

TheLonelyDevil9:
I say these kind of fixes can be merged directly @platberlitz

PR #664, opened 2026-07-12
fix(ios): add exception to viewport fix on message box

platberlitz:
This PR https://github.com/platberlitz/SillyBunny/pull/658 resolved the viewport constantly moving up in iOS WebKit. However, an exception must be made so that the message bar you use to chat doesn't keep jumping up and doing the same fix that must be done on other areas like the menu.

In essence, this is supposed to keep the previous behaviour for the message box only in iOS WebKit.

## Fix
- Stopped the iPhone chat box from jumping upward when the keyboard opens.
- Kept the keyboard fix working in menus and settings.
- Added a test to prevent the issue returning.


platberlitz:
Tested and works.

TheLonelyDevil9 (approved):

PR #665, opened 2026-07-12
fix(mobile): remove icon overlap on message box

platberlitz:
In mobile, icons always seem to overlap on the message box. This PR addresses that.

## Fix
- Added more space between the message box and the icons on both sides.
- Kept all existing icons visible and functional.
- Added a test to prevent the overlap from returning.

### Before
image

### After
image


TheLonelyDevil9 (approved):
Reviewed the implementation and regression coverage; no remaining merge blockers found.

Issue #666, opened 2026-07-14
TTS tag blocks and Dialogue Colors


dracusmorlin:
I'm using TTS Provider Kokoro. When using TTS it narrates tags like these which can be annoying and i know there is a option to disable them in the TTS setting but there's also an issue with Dialogue Colors tag:

< dialogue >
< narration >
< details >
< summary >
and since I'm using Dialogue Colors, it also reads out < font color="#whatever color" >"The dialogue here"< /font >

but if i turn on Skip < tagged > blocks, it will cause the TTS narrations to skip "quotes" dialogue because the message comes out like this and thinks it's part of the tag when using it with Dialogue Colors: 
< font color="#c8a86e" >"More water. That fire's dying. Move, girl, move."< /font >

[COLORS:Mátrina=#c8a86e,Oldřich=#b5654a,Havel=#8c9e6b]

So it's either listen to the tags being read out when using TTS or turn on Skip < tagged > blocks and let it only narrate the non quoted text and not the quoted dialogue

platberlitz:
Hello, could you please switch to the branch `staging` to check if this fixes your issue? Thank you.

PR #667, opened 2026-07-14
fix(tts): TTS extension only considers actual quotes

platberlitz:
In Issue https://github.com/platberlitz/SillyBunny/issues/666, when the the quoted dialogue has font color tags, it ignores what's actually inside.

## Fix
- Fixed TTS so font color tags no longer hide quoted dialogue.
- Prevented HTML attributes like `color="#c8a86e"` from being spoken as dialogue.
- Added tests covering the reported problem.


TheLonelyDevil9 (approved):
Reviewed the final quoted-only/dialogue-only interaction fix and regression coverage; no remaining merge blockers found.

TheLonelyDevil9:
>666

You rang?

platberlitz:
> > 666
> 
> You rang?

And after that is 667. Really, it’s a blessing.

TheLonelyDevil9:
Damn I keep falling prey to 67

PR #681, opened 2026-07-23
feat(ica): companion note cards can stay in chat history as context + ica fixes

platberlitz:
A QoL feature that makes In Chat Agents even more customisable for power users like me. So let's say you have the World Tracker as a Companion Note (which appears in the message), you can choose to have it stay in the context as though it were also generated by the main model.

Includes general ICA bug fixes.


TheLonelyDevil9:
No blocking issues, no security concerns, defaults preserve existing behavior. I'm good to go. Adding a small commit as Fable says:

> One recommendation (non-blocking): the repo has no .gitattributes eol rules, so CRLF checkouts break source-string test assertions — worth adding * text=auto eol=lf in a follow-up.

Geechan:
Check against new documentation to make sure nothing breaks new conventions. This PR touches a lot of UI elements, so will need to be examined.

platberlitz:
> Check against new documentation to make sure nothing breaks new conventions. This PR touches a lot of UI elements, so will need to be examined.

Reviewed with both PRODUCT.md and DESIGN.md.

Removed the both the hide eye and the settings gear on the note cards. The buttons are more easily tappable on phones now.

public/scripts/extensions/in-chat-agents/companion/companion-runner.js
```diff
@@ -780,7 +831,54 @@ function extractTrackerTags(text) {
 
 function buildTrackerEchoGuard(tags) {
     const examples = tags.flatMap(tag => [`[${tag}|...]`, `[/${tag}]`]).join(', ');
-    return 'HARD STOP for your reply: the bracket-format tracker notes above are read-only reference information to inform the scene. Do NOT reproduce, paraphrase, update, restate, or wrap any reply content in those tracker formats. Specifically, do not emit any of: ' + examples + ' (or variations of them). Produce your normal story reply only — never inline tracker blocks of your own.';
+    return 'HARD STOP for your reply: the bracket-format tracker notes above are read-only reference information to inform the scene. Do NOT reproduce, paraphrase, update, restate, or wrap any reply content in those tracker formats. Specifically, do not emit any of: ' + examples + ' (or variations of them). Produce your normal story reply only - never inline tracker blocks of your own.';
+}
+
+function getAuxiliaryTrackerTags() {
```
Geechan, line 857: This gathers tracker tags from every retained Companion result, after which `stripAuxiliaryTrackerEchoes()` removes every matching block from the generated reply. If an inline tracker or intentional story output uses the same tag, its valid block is silently deleted before post-generation agents run. Please use generation-specific provenance or exclude tags owned by active inline trackers rather than treating tag equality as proof of an echo.

public/scripts/extensions/in-chat-agents/style.css
```diff
@@ -759,7 +764,8 @@ button.ica--card-pill--version-update {
     }
 
     .ica--quick-chip-pin,
-    .ica--quick-chip-apply {
+    .ica--quick-chip-apply,
+    .ica--quick-chip-apply-target {
         width: 30px;
```
Geechan, line 769: The newly added chosen-target action is reduced to a 30px-wide control on mobile. This does not meet the documented 44px coarse-pointer target contract in DESIGN.md. Please provide an effective hit area of at least 44px, as already done for the Companion card actions.

public/scripts/PromptManager.js
```diff
@@ -2087,6 +2108,10 @@ class PromptManager {
     }
 
     getInChatAgentTokenUsage() {
+        if (this.runtimeAgentMessages) {
```
Geechan: `runtimeAgentMessages` is initialized to an empty `MessageCollection`, so this truthy check always succeeds and the source-token fallback below is unreachable. Before runtime assembly, or when detached token counting fails, the Agents total can incorrectly display zero. Please distinguish “no runtime snapshot” from an empty valid snapshot or gate this branch on snapshot validity.

Geechan (changes requested):
This PR substantially changes prompt assembly and runtime inspection in upstream-origin files, including `public/script.js`, `public/scripts/openai.js`, `public/scripts/group-chats.js`, and `public/scripts/PromptManager.js`, but does not update the required upstream-touch ledger. Please add test-backed entries covering the divergence reason, target seam, adapter shape, rollback path, and validation.

public/scripts/extensions/in-chat-agents/companion/companion-runner.js
platberlitz: Fixed. Active tracker output is now protected from cleanup, with regression tests added.

public/scripts/extensions/in-chat-agents/style.css
platberlitz: Fixed. The button now has a 44×44px touch target on touch devices.

public/scripts/PromptManager.js
platberlitz: Fixed. Missing runtime data now uses the fallback count, while a valid empty result still shows zero.

platberlitz:
Added the requested ledger entries, including tests, validation, and rollback details.

Geechan (approved):

PR #684, opened 2026-07-24
fix(lorebooks): general lorebook bug fixes

platberlitz:
This PR audits lorebooks for general bugs and fixes bugs that I've found, such as percentage chances not working, and excluding the name of the character doesn't include if the character card is a group card (e.g. my card Kaveh & Alhaitham constantly triggers the entries for Kaveh and Alhaitham and their relationships separately.)

### Main Problems Fixed
- Percentage chances could be ignored, especially for imported lorebooks.
- A 0% entry could rarely activate.
- Character names from greetings could trigger lore when “Include Names” was disabled.
- Group-card names like “Kaveh & Alhaitham” could trigger separate entries unintentionally.
- Empty keywords could activate entries unexpectedly.
- Some entries rejected by the lorebook budget could still affect later scans.
- Grouped lore entries could remove the wrong entry.
- Recursion, cooldowns, sticky entries, ordering, and scan-depth settings had several edge-case errors.

### Data Safety
- Editing two lorebooks quickly could lose one book’s changes.
- Failed saves could be reported as successful.
- Renaming could overwrite or delete another lorebook.
- Imports, replacements, and deletions could race against unsaved edits.
- Chat, character, and persona lorebook links could break after renaming.
- Very long and unusual lorebook names are now handled safely.
- Invalid lorebook files can no longer disable every other active lorebook.

### Import And Character Cards
- Standard CharacterBook and Lorebook V3 files can now be imported properly.
- Imported metadata and unknown fields are preserved more reliably.
- Duplicate or unusual entry IDs no longer overwrite entries.
- Regex keywords, capitalization settings, probability settings, and ordering survive import/export more accurately.
- Embedded character-card lorebooks now save and link correctly.
- Batch imports detect duplicate or equivalent names before overwriting anything.

### Editor Fixes
- Merely opening an entry no longer silently changes its settings.
- New entries are saved immediately.
- Unknown character filters are preserved.
- Rapidly switching lorebooks no longer displays the wrong book.
- Search, sorting, group weights, and recursion-level controls behave more consistently.

### DIFF
--- a/public/scripts/world-info-character-book.js
+++ b/public/scripts/world-info-character-book.js
@@ -150,3 +150,71 @@ export function serializeCharacterBookKeys(primaryKeys = [], secondaryKeys = [])
         useRegex,
     };
 }
+
+/**
+ * Serializes a native World Info entry into a CharacterBook-format record.
+ *
+ * Mirrors the per-entry output of the server-side character book converter
+ * (src/endpoints/characters.js, convertWorldInfoToCharacterBook) so records
+ * appended to a book's originalData are complete rather than lossy. Legacy
+ * positions (e.g. the string '0') are normalized before choosing the
+ * before_char/after_char label; anything that does not normalize to
+ * `positions.before` serializes as 'after_char', matching server semantics.
+ *
+ * @param {object} entry Native World Info entry
+ * @param {{ before: number, after: number, ANTop: number, ANBottom: number, atDepth: number, EMTop: number, EMBottom: number, outlet: number }} positions World Info position enum
+ * @returns {object} CharacterBook-format entry record
+ */
+export function serializeWorldInfoEntry(entry, positions) {
+    const { keys, secondaryKeys, useRegex } = serializeCharacterBookKeys(entry.key ?? [], entry.keysecondary ?? []);
+    const normalizedPosition = normalizeWorldInfoPosition(entry.position, positions);
+    return {
+        id: entry.uid,
+        keys,
+        secondary_keys: secondaryKeys,
+        comment: entry.comment ?? '',
+        content: entry.content ?? '',
+        constant: entry.constant ?? false,
+        selective: entry.selective ?? false,
+        insertion_order: entry.order ?? 100,
+        enabled: !entry.disable,
+        position: normalizedPosition === positions.before ? 'before_char' : 'after_char',
+        use_regex: useRegex,
+        case_sensitive: entry.caseSensitive ?? null,
+        extensions: {
+            ...entry.extensions,
+            position: normalizedPosition ?? entry.position,
+            exclude_recursion: entry.excludeRecursion ?? false,
+            display_index: entry.displayIndex,
+            probability: entry.probability ?? null,
+            useProbability: entry.useProbability ?? false,
+            depth: entry.depth ?? 4,
+            selectiveLogic: entry.selectiveLogic ?? 0,
+            outlet_name: entry.outletName ?? '',
+            group: entry.group ?? '',
+            group_override: entry.groupOverride ?? false,
+            group_weight: entry.groupWeight ?? null,
+            prevent_recursion: entry.preventRecursion ?? false,
+            delay_until_recursion: entry.delayUntilRecursion ?? false,
+            scan_depth: entry.scanDepth ?? null,
+            match_whole_words: entry.matchWholeWords ?? null,
+            use_group_scoring: entry.useGroupScoring ?? false,
+            case_sensitive: entry.caseSensitive ?? null,
+            automation_id: entry.automationId ?? '',
+            role: entry.role ?? 0,
+            vectorized: entry.vectorized ?? false,
+            sticky: entry.sticky ?? null,
+            cooldown: entry.cooldown ?? null,
+            delay: entry.delay ?? null,
+            match_persona_description: entry.matchPersonaDescription ?? false,
+            match_character_description: entry.matchCharacterDescription ?? false,
+            match_character_personality: entry.matchCharacterPersonality ?? false,
+            match_character_depth_prompt: entry.matchCharacterDepthPrompt ?? false,
+            match_scenario: entry.matchScenario ?? false,
+            match_creator_notes: entry.matchCreatorNotes ?? false,
+            triggers: entry.triggers ?? [],
+            ignore_budget: entry.ignoreBudget ?? false,
+            agent_blacklisted: entry.agentBlacklisted ?? false,
+        },
+    };
+}
--- a/public/scripts/world-info-character-book.js
+++ b/public/scripts/world-info-character-book.js
@@ -77,5 +77,76 @@ export function normalizeWorldInfoPosition(position, positions) {
 export function normalizeCharacterBookPosition(extensionPosition, entryPosition, positions) {
     return normalizeWorldInfoPosition(extensionPosition, positions)
         ?? normalizeWorldInfoPosition(entryPosition, positions)
-        ?? positions.after;
+        ?? positions.before;
+}
+
+/**
+ * Escapes regex delimiter slashes without adding duplicate escaping on repeated conversions.
+ * @param {string} value Raw CharacterBook regex
+ * @returns {string} Regex body safe for slash-delimited storage
+ */
+export function escapeCharacterBookRegex(value) {
+    return String(value).replace(/(\\*)\//g, (_match, backslashes) => {
+        const delimiterEscape = backslashes.length % 2 === 0 ? '\\' : '';
+        return `${backslashes}${delimiterEscape}/`;
+    });
+}
+
+function unescapeCharacterBookRegex(value) {
+    return String(value).replace(/(\\*)\//g, (_match, backslashes) => {
+        const rawBackslashes = backslashes.length % 2 === 1 ? backslashes.slice(0, -1) : backslashes;
+        return `${rawBackslashes}/`;
+    });
+}
+
+function parseWorldInfoRegexKey(value) {
+    const match = String(value).match(/^\/([\s\S]+)\/([gimsuy]*)$/);
+    if (!match) {
+        return null;
+    }
+
+    const body = match[1];
+    for (let index = 0; index < body.length; index++) {
+        if (body[index] !== '/') {
+            continue;
+        }
+        let backslashes = 0;
+        for (let cursor = index - 1; cursor >= 0 && body[cursor] === '\\'; cursor--) {
+            backslashes++;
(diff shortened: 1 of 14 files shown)


public/scripts/world-info-character-book.js
```diff
@@ -77,5 +77,76 @@ export function normalizeWorldInfoPosition(position, positions) {
 export function normalizeCharacterBookPosition(extensionPosition, entryPosition, positions) {
     return normalizeWorldInfoPosition(extensionPosition, positions)
         ?? normalizeWorldInfoPosition(entryPosition, positions)
-        ?? positions.after;
```
Geechan, line 80: What's the rationale for changing the default position from `positions.after` to `positions.before` (and so on throughout the commit history)? Unless there's a good reason for it, this seems like it will break the default established positioning as found in upstream SillyTavern. The rest of the changes seem like they don't require this position change.

This is also not documented on the PR.

platberlitz: Regression resolved. Let me know if you find anything else

public/scripts/world-info.js
```diff
@@ -4385,20 +4488,116 @@ export function createWorldInfoEntry(_name, data) {
...
 }
 
+function appendWIOriginalDataEntry(data, entry) {
+    if (!data.originalData || !Array.isArray(data.originalData.entries)) {
+        return;
+    }
+
+    data.originalDataUidMap ??= {};
+    data.originalDataUidMap[entry.uid] = data.originalData.entries.length;
+    data.originalData.entries.push({
```
Geechan: appendWIOriginalDataEntry only copies a subset of the native entry and hard-codes use_regex: false. When a native entry is moved into a CharacterBook-backed lorebook, sourceOriginalEntry is unavailable, so exporting that lorebook into a character card loses regex, recursion, group, matching, sticky/cooldown, and other extension fields.

This is also found in [line 6824](https://github.com/SillyBunnyTeam/SillyBunny/pull/684/changes#diff-657e264183d3b1542514cd29f037b6f3545f14e673dd29b7491999a849ce8496R6824)

public/scripts/world-info.js
```diff
@@ -4385,20 +4488,116 @@ export function createWorldInfoEntry(_name, data) {
...
+        id: entry.uid,
+        keys: structuredClone(entry.key ?? []),
+        secondary_keys: structuredClone(entry.keysecondary ?? []),
+        comment: entry.comment ?? '',
+        content: entry.content ?? '',
+        constant: entry.constant ?? false,
+        selective: entry.selective ?? false,
+        insertion_order: entry.order ?? 100,
+        enabled: !entry.disable,
+        position: entry.position === world_info_position.before ? 'before_char' : 'after_char',
```
Geechan: This is still a strict comparison, unlike the corrected conversion paths. A legacy native entry with position '0' moved into an imported CharacterBook is serialized as after_char, so the position regression is still present in this move/create path.

public/scripts/world-info.js
```diff
@@ -5207,15 +5519,21 @@ export async function checkWorldInfo(chat, maxContext, isDryRun, globalScanData
         console.debug(`[WI] Search done. Found ${activatedNow.size} possible entries.`);
 
         // Sort the entries for the probability and the budget limit checks
+        const entryOrder = new Map(sortedEntries.map((entry, index) => [getWorldInfoEntryKey(entry), index]));
         const newEntries = [...activatedNow]
             .sort((a, b) => {
                 const isASticky = timedEffects.isEffectActive('sticky', a) ? 1 : 0;
                 const isBSticky = timedEffects.isEffectActive('sticky', b) ? 1 : 0;
-                return isBSticky - isASticky || sortedEntries.indexOf(a) - sortedEntries.indexOf(b);
+                const isAConstant = a.constant ? 1 : 0;
```
Geechan: Upstream prioritizes sticky entries, then preserves the configured insertion order. Promoting every constant entry changes budget selection: a low-priority constant can displace an earlier keyed entry when the lorebook budget is constrained. Is this ordering change intentional and tested?

public/scripts/world-info.js
```diff
@@ -620,7 +628,7 @@ class WorldInfoTimedEffects {
         return {
             hash: this.#getEntryHash(entry),
             start: this.#chat.length,
-            end: this.#chat.length + Number(entry[type]),
+            end: this.#chat.length + Number(entry[type]) + (isProtected ? 0 : 1),
```
Geechan: This changes the established expiry from chat.length + duration to chat.length + duration + 1 for non-protected effects. Since expiration checks use chat.length >= end, sticky/cooldown effects now remain active for one additional message. Please add a focused duration test or retain upstream semantics.

src/endpoints/worldinfo.js
```diff
@@ -98,36 +213,36 @@ router.post('/delete', (request, response) => {
...
-    if (!worldName) {
-        return response.status(400).send('World file must have a name');
-    }
+        if (!worldName) {
+            return response.status(400).send('World file must have a name');
+        }
 
-    tryWriteFileSync(pathToNewFile, fileContents);
-    return response.send({ name: worldName });
+        writeWorldInfoFile(pathToNewFile, fileContents);
```
Geechan, line 234: This catch handles writeWorldInfoFile() failures as invalid-file errors and returns HTTP 400. Disk-full, permission, and rollback failures are server/storage failures, not malformed input; they should retain 5xx behavior like the edit and rename routes.

public/scripts/world-info.js
```diff
@@ -5010,7 +5323,8 @@ export async function checkWorldInfo(chat, maxContext, isDryRun, globalScanData
             };
 
             // Already processed, considered and then skipped entries should still be skipped
-            if (failedProbabilityChecks.has(entry) || allActivatedEntries.has(`${entry.world}.${entry.uid}`)) {
+            const entryKey = getWorldInfoEntryKey(entry);
```
Geechan: Using ${world}.${uid} as the identity means two native entries with distinct object keys but duplicate embedded UIDs share state. If one fails probability, the other can be skipped on recursion; if both activate, the later entry overwrites the activation map entry. Duplicate embedded UIDs are not rejected by native World Info validation.

Related action overwrite: [public/scripts/world-info.js:5592 ](https://github.com/SillyBunnyTeam/SillyBunny/pull/684/changes#diff-657e264183d3b1542514cd29f037b6f3545f14e673dd29b7491999a849ce8496R5592)

Geechan (changes requested):

public/scripts/world-info.js
platberlitz: Fixed in 612ea61d6. Background: when a lorebook was imported from a character card, the app keeps a hidden copy of it in the card's original format, and that hidden copy is what gets written back into the card on export. The bug: when you created, moved, or duplicated an entry into such a lorebook, the hidden copy only recorded the basic fields - regex keywords, recursion options, groups, match settings, sticky/cooldown timers and the rest were left out, so exporting the book into a card silently lost them. 

There is now one shared "translator" (`serializeWorldInfoEntry` in world-info-character-book.js) that writes every field into the card format, and all three paths (create, move, duplicate) use it. The server-side exporter also gained the one field it was missing (`agent_blacklisted`), so both sides now write the same complete record.

public/scripts/world-info.js
platberlitz: Fixed in the newest commit

Some older lorebooks store the entry position as text, the string '0' instead of the number 0. The previous code only recognized the exact number, so those entries were recorded as "after character" even though '0' means "before character". The new translator first converts any legacy form ('0', 'before_char', and so on) into the proper value before deciding, so the position survives correctly on create, move, and duplicate. There is a regression test for exactly this case.

public/scripts/world-info.js
platberlitz: Reverted in the newest commit

Selection order is back to exactly what upstream does: sticky entries first, then the insertion order the user configured. The promoted-constants version could let a low-priority constant entry eat the token budget ahead of entries the user had ranked higher. Anyone who needs a constant entry to always make it in under a tight budget can use the existing "ignore budget" option on that entry. A side benefit of the revert: entries force-activated by extensions go back to the front, which also matches upstream.

public/scripts/world-info.js
platberlitz: Went with the "add a focused duration test" option. 612ea61d6 moves the calculation into a small standalone helper (`getTimedEffectWindow`) with tests that pin the behavior down. Why I would rather keep the change than revert with the old formula:

sticky 1 and cooldown 1 did nothing at all, the effect expired before it could ever apply. The old behavior was also inconsistent with itself: a cooldown that begins when a sticky runs out lasted the full N messages, while a normal cooldown only lasted N-1. With this change, a duration of N means N messages everywhere, and a duration of 1 actually works. The one path that deliberately keeps the old formula (the cooldown that follows a sticky) already counts the current message, which the helper's comment and a dedicated test now explain.

src/endpoints/worldinfo.js
platberlitz: Fixed in 612ea61d6. The import endpoint now distinguishes "the file you sent is bad" from "the server failed to save it", the same way the edit and rename endpoints do.

Broken JSON, a file that is not a lorebook, or a name that comes out empty → 400.
The disk being full, missing permissions, or the safety-rollback failing → logged on the server and returned as 500. One small wording change: an unusable name now gets the message "World file must have a name" instead of the generic "not a valid file" text. New tests cover the success paths, each of the 400s, and a real write failure that must come back as 500.

public/scripts/world-info.js
platberlitz: Addressed in 612ea61d6. The regression:

This PR started remembering failed probability rolls by the entry's "book name + ID" label, so two entries that end up with the same ID (through a hand-edited or third-party file) also shared the failure - one entry's bad roll silenced the other for the rest of the scan. That tracking is back to remembering the actual entry, as upstream does, and the sort lookup likewise, so same IDs no longer affect each other's rolls or ordering. What I intentionally did not change: the list of already-activated entries is still keyed by that label, so once one twin activates the other is skipped, but that is exactly upstream's current behavior, and changing it would break the force-activate feature used by extensions, which relies on matching entries by that same label. The real cure would be making IDs genuinely unique when a file is loaded (or rejecting duplicates outright), but that would invalidate saved sticky/cooldown timers for affected books, so I would rather do it as a separate follow-up if you think it is worth it.

Geechan (approved):
With all current changes, this PR is good to go. All regressions and upstream incompatibilities have been addressed.

PR #685, opened 2026-07-25
feat(ui): add option called “Icons only top bar” to put every icon on the top bar

platberlitz:
An option in the Settings page called "Icons only top bar" to remove the text and place every page icon on the top bar instead, similar to upstream SillyTavern.
Includes any user-toggled Quick Actions buttons.
Note: Workspace buttons always on the left regardless of custom Quick Actions
Reviewed against the new PRODUCT.md and DESIGN.md.

Looks like this:
1. Tablet size
image

2, Desktop size
image

3. Mobile - Scrollable horizontally (Btw there is a divider, just not visible with my personal theme)
image

Turn it on here
image

**Note**: this technically goes against one of the philosophies in DESIGN.md: "Don't hide essential writing controls behind mystery meat navigation." The icons-only mode is, by definition, label-free. Each button has a title attribute and the cluster grouping provides spatial context, but on touch devices title is not surfaced at all. 

However, given this is opt-in and a deliberate design decision, it is safe to implement with that concession.

### DIFF
--- a/docs/upstream-touch-ledger.md
+++ b/docs/upstream-touch-ledger.md
@@ -92,14 +92,14 @@ This ledger tracks intentional SillyBunny divergence in upstream-origin files. I
 ### `public/scripts/sillybunny-tabs.js` - shell chat controls
 | Field | Value |
 | --- | --- |
-| Area | Mobile shell, chat navigation, and preset/API sync. |
-| Divergence reason | SillyBunny shell owns top/bottom navigation, chat controls, drawers, mobile actions, search shortcut focus, viewport-reset dispatches, and mirrored connection-profile controls that interact with chat and API state. |
+| Area | Mobile shell, top-bar layout, chat navigation, and preset/API sync. |
+| Divergence reason | SillyBunny shell owns top/bottom navigation, chat controls, drawers, mobile actions, search shortcut focus, viewport-reset dispatches, and mirrored connection-profile controls that interact with chat and API state. Its per-device icons-only top bar adds canonical Workspace, Customize, and Characters page clusters, hides redundant shell anchors, deduplicates Quick Access targets, and keeps trailing controls fixed while the page rail pans. |
 | Target seam | `public/scripts/chat-render-lifecycle/` for chat scroll requests; `public/scripts/mobile-shell-lifecycle/` for drawer/nav/viewport behavior; `public/scripts/preset-api-sync-lifecycle/` for active API and connection-profile mirror decisions. |
-| Adapter shape | Shell code keeps DOM wiring and requests lifecycle decisions for drawer bounds, viewport sync order, drawer-bound scheduling, overlay exclusivity, rail quick-action normalization and visibility, inline drawer auto-close and persistence keys, nav drag, page scroll, overlay open/close, auto-close, modal inert policy, search shortcut pre-focus, viewport reset timing, active API connect-button lookup, connection-profile source binding, connection-profile source mutation rebind decisions, connection-profile mirror state, connection-profile mirror updates, connection-profile status text, and connection-strip open state. |
-| Protecting tests | `tests/mobile-shell-lifecycle.test.js`, `tests/mobile-shell-lifecycle-drawer-bounds.test.js`, `tests/mobile-shell-lifecycle-viewport-sync.test.js`, `tests/mobile-shell-lifecycle-overlay-exclusion.test.js`, `tests/mobile-shell-lifecycle-rail-model.test.js`, `tests/mobile-shell-lifecycle-inline-drawers.test.js`, `tests/mobile-shell-lifecycle-wiring.test.js`, `tests/mobile-shell-smoke.e2e.js`, `tests/preset-api-sync-lifecycle.test.js`, `tests/preset-api-sync-lifecycle-wiring.test.js`, future shell smoke checks for drawer/tab/preset/chat-scroll behavior. |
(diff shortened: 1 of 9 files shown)


Geechan:
Where does one find the option to toggle this mode?

I'm afraid to say it, but this looks very ugly IMO, and even as a non default, it feels sub-optimal judging from the screenshots. There's so much visual clutter that it becomes very difficult to discern what each button does. We're effectively dumping all the sub menu buttons in one place to the detriment of the UI's simplicity, which goes against our new PRODUCT.md specifications.

If implementing an icon-only mode (which is not inherently a bad idea), it should still respect the layer system detailed in PRODUCT.md. The top bar should have a set amount of elements in it at all times, to avoid this ridiculous visual clutter. Comparing to SillyTavern's top bar: this configuration doesn't account for other elements we've added, such as the title bar, home bar, search, etc. which all contribute to make the experience entirely inconsistent compared to upstream's implementation.

If a compromise can't be made, then this might be better suited as an optional extension given it goes against the project's design philosophies so strongly.

platberlitz:
> Where does one find the option to toggle this mode?
> 
> I'm afraid to say it, but this looks very ugly IMO, and even as a non default, it feels sub-optimal judging from the screenshots. There's so much visual clutter that it becomes very difficult to discern what each button does. We're effectively dumping all the sub menu buttons in one place to the detriment of the UI's simplicity, which goes against our new PRODUCT.md specifications.
> 
> If implementing an icon-only mode (which is not inherently a bad idea), it should still respect the layer system detailed in PRODUCT.md. The top bar should have a set amount of elements in it at all times, to avoid this ridiculous visual clutter. Comparing to SillyTavern's top bar: this configuration doesn't account for other elements such as the title bar, home bar, search, etc. which all contribute to make the experience entirely inconsistent compared to upstream's implementation.
> 
> If a compromise can't be made, then this might be better suited as an optional extension given it goes against the project's design philosophies so strongly.

image

Completely optional.

I disagree with creating an extension just for this when an optional toggle would have you, with intent, turn it on. I agree I personally don't want to use this all the time myself, but there's essentially nothing to maintain.

You can suggest where to place which and where instead. I don't really want to remove this option as I've been pinged enough times about this particular complaint.

Geechan:
> I disagree with creating an extension just for this when an optional toggle would have you, with intent, turn it on. I agree I personally don't want to use this all the time myself, but there's essentially nothing to maintain.

> You can suggest where to place which and where instead. I don't really want to remove this option as I've been pinged enough times about this particular complaint.

That's fine, but I have listed some possible solutions in my original comment if we wish to keep it as an option. Maintainability is not the concern here: consistency is.

If implementing a legacy icon-only mode, visual clutter should be reduced by limiting the amount of options there to only what's needed. As I said: the issue copying SillyTavern's icon-only design is it doesn't account for all our added content that each does something different from each other. Without being able to distinguish them with padding, general sizing, and text, it becomes more cluttered.

Additionally, the different buttons should be placed in more appropriate locations, with some appropriate gaps. At the moment it's a bit of a mess. 

Here's some ideas:

Far left side: all buttons pertaining to Workspace
Left side: all buttons pertaining to Customize

Titlebar

Right side: home button, quick action button
Far right side: all buttons pertaining to Characters menu

This better delegates all the buttons into sensible locations. Additionally, the buttons should be larger and more distinct so they are more clearly visible. They are currently small targets that are difficult to select.

platberlitz:
> > I disagree with creating an extension just for this when an optional toggle would have you, with intent, turn it on. I agree I personally don't want to use this all the time myself, but there's essentially nothing to maintain.
> 
> > You can suggest where to place which and where instead. I don't really want to remove this option as I've been pinged enough times about this particular complaint.
> 
> That's fine, but I have listed some possible solutions in my original comment if we wish to keep it as an option. Maintainability is not the concern here: consistency is.
> 
> If implementing a legacy icon-only mode, visual clutter should be reduced by limiting the amount of options there to only what's needed. As I said: the issue copying SillyTavern's icon-only design is it doesn't account for all our added content that each does something different from each other. Without being able to distinguish them with padding, general sizing, and text, it becomes more cluttered.
> 
> Additionally, the different buttons should be placed in more appropriate locations, with some appropriate gaps. At the moment it's a bit of a mess.
> 
> Here's some ideas:
> 
> Far left side: all buttons pertaining to Workspace Left side: all buttons pertaining to Customize
> 
> Titlebar
> 
> Right side: home button, quick action button Far right side: all buttons pertaining to Characters menu
> 
> This better delegates all the buttons into sensible locations. Additionally, the buttons should be larger and more distinct so they are more clearly visible. They are currently small targets that are difficult to select.

That makes more sense to me. Will implement that and update the PR body after.

Geechan:
Another general note: the positioning for the toggle is a bit weird IMO. With the new PRODUCT.md file and design language, I've actually discouraged the use of nested menus beyond the first one found in layer 4 (the main configuration layer), so going forward, it would be prudent to try and find a better location for it. For the time being, place the toggle in the Desktop (and mobile) navigation category. This is where I would intuitively expect it.
image

Geechan:
Remove the extraneous workspace and customize buttons as these should be covered by the rest of the corresponding icons. Re-add formatting, agents, server, console logs, editor, import buttons to their respective positions in the top bar. For some reason these are missing now.

Otherwise, positioning looks solid.

Geechan:
Looking excellent and very polished now. One more thing needs to be addressed: add information in docs/upstream-touch-ledger.md to reflect this change, as it adds diversions from upstream SillyTavern.

This PR modifies three files with existing ledger entries:

public/scripts/sillybunny-tabs.js — two entries, both with Last reviewed dates of 2026-06-02 and 2026-06-13. The new icons-only divergence (new functions, modified buildTopBar, getShellProxyButton, stopProxyPointerPropagation, forceDrawerState, syncCharacterShellTabs, syncCharacterDrawerStateFromDom, setActiveTab, updateTopBarBrand, updateShortcutButton, initAll) must be recorded. Protecting tests should add topbar-label-tap-cycle.test.js.

public/css/sillybunny-mobile-shell.css — part of the "menu polish assets" entry, Last reviewed 2026-06-02.

public/index.html — version string bumped, covered by same entry.

Add these to docs/upstream-touch-ledger.md, and this will be good to merge.

Geechan:
Additionally, the character menu icon highlights itself even when selecting another item in the character subsection. It also does not explicitly open the character menu - this should be addressed so it explicitly opens the Characters sub-menu.

Make sure this change does not break the non icon layout too - it should be isolated to this mode specifically.

Geechan:
New regression: when deselecting a menu by clicking on the main window, removing it from focus, the icon for the last previously selected icon stays active. This does not apply for the character menu sub-sections.

image

Geechan (approved):
All issues addressed. Ready to merge now.

PR #689, opened 2026-07-26
feat(launcher): gate bun --smol behind SILLYBUNNY_BUN_SMOL

platberlitz:
Adds bun --smol as an argument for start.sh.

```SILLYBUNNY_BUN_SMOL=1 ./start.sh```

Used to start for low-memory devices or devices prone to memory leaks that use pm2, docker or Termux (which use start.sh rather than bun run, which can do bun run --smol, but not pm2 or docker).

Added to the readme.md about it myself for clarification

### DIFF
--- a/.github/readme.md
+++ b/.github/readme.md
@@ -110,13 +110,15 @@ bun run start:global   # SillyBunny-owned data paths
 bun run start:no-csrf  # disable CSRF (local dev)
 ```
 
-When you start through a launcher script through pm2, systemd or Docker, rather than `bun run`, set `SILLYBUNNY_BUN_SMOL=1` to get the same `bun run --smol` low-memory mode:
+If you launch through a script rather than `bun run` — pm2, systemd, Docker, or Termux — set `SILLYBUNNY_BUN_SMOL=1` to get the same low-memory mode as `start:mobile`:
 
 ```bash
 SILLYBUNNY_BUN_SMOL=1 ./start.sh
 ```
 
-This flag isn't supported in Node.js mode.
+`start.sh`, `Start.bat`, and the Docker entrypoint all honour it; for Docker, set it under `environment:` in `docker/docker-compose.yml`.
+
+`--smol` is a Bun flag, so it does nothing in Node.js mode — and Termux, macOS, and ARM hosts pick Node.js by default. Use `./start-bun.sh` (Termux: `bash start-termux-bun.sh`) to get Bun together with `--smol`. The launcher warns when the flag is set while Node.js is selected.
 
 ### macOS notes
 
--- a/.github/readme.md
+++ b/.github/readme.md
@@ -110,6 +110,14 @@ bun run start:global   # SillyBunny-owned data paths
 bun run start:no-csrf  # disable CSRF (local dev)
 ```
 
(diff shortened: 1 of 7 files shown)


TheLonelyDevil9:
Made some changes.

Pushed a follow-up commit (`ac319c96c`) onto this branch. Fixes outlined below:

**Docker didn't work.** The README said pm2, systemd and Docker all pick the flag up, but Docker never runs `start.sh` — the image entrypoint is `docker-entrypoint.sh`, which launched `bun server.js` unconditionally. Setting `SILLYBUNNY_BUN_SMOL=1` under `environment:` in Compose silently did nothing, and a container memory cap is the case this flag exists for. The entrypoint now gates `--smol` behind an `is_truthy` mirroring the one in `start.sh`, and `docker/docker-compose.yml` carries a commented `SILLYBUNNY_BUN_SMOL=1` line so Compose users can find it.

**The flag was inert on Termux, the motivating case.** `prefer_node_runtime()` picks Node on Termux unless `SILLYBUNNY_TERMUX_RUNTIME=bun` is set, and Node is also the default on macOS and ARM. So `SILLYBUNNY_BUN_SMOL=1 bash start.sh`, the command in the PR description, was a no-op there. "This flag isn't supported in Node.js mode" was accurate but never connected to "these platforms default to Node," leaving no way to tell that the flag you just set was doing nothing. Both launchers now warn in that case, matching the `[SillyBunny] …` lines `start.sh` already prints for its macOS/ARM runtime overrides, and the README explains how to reach Bun on those platforms.

**`Start.bat` accepted only `1`.** `is_truthy` matches `1`/`true`/`yes`/`on` case-insensitively, so `SILLYBUNNY_BUN_SMOL=true` worked on Linux and silently didn't on Windows. Normalised to a `_bun_smol` 0/1 flag covering all four spellings, following the existing `_auto_update` pattern at the top of the file.

**The new tests failed on Windows.** `execFileSync('bash', …)` assumes `bash` on `PATH` is a usable POSIX bash; on Windows it resolves to the WSL app-execution alias ahead of Git Bash, which fails outright when no distro is installed. Five of seven tests failed locally while CI stayed green, since `run-tests` only runs on `ubuntu-latest`. The suite now probes whether `bash` can actually run a script in the temp dir and skips the shell block when it can't. Parity assertions were extended to cover the Docker entrypoint and both Node-mode warnings.

Also rewrote the README paragraph (it read "start through a launcher script through pm2") and regenerated `.github/readme.md` with `scripts/sync-readme-mirror.sh` rather than editing it by hand.

Neither Docker nor `Start.bat` is covered by CI, so I exercised both rather than grepping the source:

- jest with a working bash: 12/12 pass. On Windows with the WSL alias: 7 pass, 5 skipped, suite green.
- `Start.bat` tail driven through `cmd.exe`: 20 runtime × value combinations dispatch correctly, warning included.
- `docker-entrypoint.sh` `start_sillybunny()` executed under `sh` with a stubbed `bun`: 13/13 cases.
- `start.sh` Node-mode warning: 8 combinations, fires only on truthy + Node.
- `bash -n` / `sh -n` clean, eslint clean, Compose YAML parses with the commented line correctly excluded, README mirror `--check` passes.

PR #696, opened 2026-07-26
chore: Add labels for 23 control elements

TheLonelyDevil9:
API, source, model selects, custom API key, sample sliders, counters, seed, logit bias preset, etc are getting a11y control labels.


TheLonelyDevil9:
I have a feeling this is related to your recent removal of labels or something @platberlitz, so if it seems like it, I'll close it

PR #697, opened 2026-07-26
chore: hardcoded values revised for z index in extension menu/options/popper modal

TheLonelyDevil9:
Technical details:

`style.css`

`#options`/`#extensionsMenu`/`.popper-modal` used a hardcoded `z-index: 29999` outside the `--sb-z-*` scale. Now `var(--sb-z-critical, 10000)` — verified nothing legitimate sits between 10000 and 29999, so stacking order is unchanged (zoomed avatars at 9999 stay beneath).

### DIFF
--- a/public/style.css
+++ b/public/style.css
@@ -1260,7 +1260,9 @@ body .panelControlBar {
 #extensionsMenu,
 .popup .popper-modal {
     display: flex;
-    z-index: 29999;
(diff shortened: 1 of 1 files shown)


TheLonelyDevil9:
Honestly just feels like basic housekeeping, let me know if it adds value or we just close it as-is.

PR #703, opened 2026-07-29
feat: reuse Start Reply With for Kimi K3 partial prefill

platberlitz:
Kimi K3 supports Partial Prefill, and it already considers it so when using Start Reply With, so just reusing that an exposing it in the UI. Parameter ends up being partial: true, by default.

Under Advanced & Reasoning:
image

For NanoGPT, Moonshot, OpenRouter and Custom backends.
Also, Advanced Formatting has Start Reply With which works too.

Gated to only Kimi K3 on all relevant backends.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2091,7 +2091,7 @@
                                             </div>
                                         </div>
                                     </div>
-                                    <div class="range-block" data-source="custom,moonshot,nanogpt">
+                                    <div class="range-block" data-source="custom,moonshot,nanogpt,openrouter">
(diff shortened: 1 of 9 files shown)


Geechan:
I limited the scope of this to only specifically target K3, as it seems to be a feature that only works with that model, anyway. If that's a problem, let me know.

Otherwise, good to merge.

platberlitz:
Fine with me.

Geechan (approved):

platberlitz:
I forgot OpenRouter so... yeah. Added that.

PR #726, opened 2026-08-04
docs: add Conversation Mode glossary, rest API docs, and ICA glossary

platberlitz:
Adds these docs to be referred to separately from the main readme. Double-checked personally with my own extensive use case and knowledge, and me implementing these, and checking with the LLM for any hallucinations or inaccuracies.

Edit as needed if you'd like before merging.


Geechan:
Removed READMEs while keeping the glossaries and API documentation intact. The READMEs were far too verbose for an end user, and are better suited to an eventual Wiki page as a part of extended documentation (or in-app documentation).

Glossaries are still useful for contributor and agent use, so they can stay.

platberlitz:
> Removed READMEs while keeping the glossaries and API documentation intact. The READMEs were far too verbose for an end user, and are better suited to an eventual Wiki page as a part of extended documentation (or in-app documentation).
> 
> Glossaries are still useful for contributor and agent use, so they can stay.

Approved of the removal.

PR #727, opened 2026-08-04
fix(world_info): fix certain values being simplified, instead it now leaves values verbatim

platberlitz:
An upstream SillyTavern bug makes some specific features not work due to how it simplifies values. 00 comes back as 0, 007 as 7, 0.50 as 0.5. This is normal for regular use case, but it causes compatibility errors with https://github.com/SillyBunnyTeam/SillyBunny-MacroEnhanced which uses more advanced macros. The time/clock macro would show up as 7:0 PM instead of 7:00 PM for example.

This fix makes the numeric values be rendered verbatim rather than simplified.

Essentially this is a compatibility fix to help macro expansions without changing anything huge. Could be beneficial in fact to keep it verbatim.

For reviewers: Seeking if this is low-risk as it fixes an upstream ST bug.

### DIFF
--- a/public/scripts/slash-commands/SlashCommandRuntimeUtils.js
+++ b/public/scripts/slash-commands/SlashCommandRuntimeUtils.js
@@ -38,6 +38,41 @@ export function readVariableValue(value) {
     return Number(value);
 }
 
+/**
+ * Whether an operand can take part in a numeric comparison.
+ *
+ * @param {any} value
+ * @returns {boolean}
+ */
+export function isNumericOperand(value) {
+    if (typeof value === 'number') {
+        return true;
+    }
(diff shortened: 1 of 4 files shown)


Geechan:
Latest commit addresses some upstream incompatibilities while trying to keep compatibility with the enhanced macro extension working. Please check.

platberlitz:
> Latest commit addresses some upstream incompatibilities while trying to keep compatibility with the enhanced macro extension working. Please check.

Checked commit against upstream compatibility, was correct, along with compatibility with the extension. Pushed a commit for the test coverage which was missing `evalBoolean`.

Should be good to merge.

Geechan:
Added a missing SB divergence comment. Now we're good to merge.

PR #729, opened 2026-08-04
fix: fix CharacterLibrary and ST-CoPilot incompatibilities

platberlitz:
## CharacterLibrary
Third-party extension [CharacterLibrary](https://github.com/Sillyanonymous/SillyTavern-CharacterLibrary) is a well-used extension, this PR involves making SB have full parity with it. Addresses the specific named issue: "If in embedded panel mode (possibly not just then) it seems to take over the entire top bar and blocks all of the other menus if you forget to turn on "Show launcher dropdown on Characters button". Easy enough to fix by turning it on, but I had to use the inspect console to dike out its button in order to reach the settings menu again."
Audits for other compatibility patches as well.

- Extensions adding a plain (non-drawer) button were getting one that was completely unclickable.
- CharacterLibrary's chevron marker was being attached to a hidden copy of the Characters button, so you never saw that the button had a menu. It's now mirrored onto the visible one.
- Its embedded panel was anchored to the height of only the top row, so it covered the chat bar row underneath. Now it's anchored to the whole bar.
- Third-party buttons in the composer were hidden outright on phones, and third-party settings panels were invisible in all four settings tabs. Both now show up.

## ST-Copilot
Another popular third-party extension, [ST-Copilot](https://github.com/Supker/ST-Copilot), also doesn't quite work correctly in SillyBunny.
SillyBunny doesn't resolve if there is a typo in reasoning effort (e.g. capital M Medium) so it fails. Should now resolve properly by automatically correcting capitalisation, and adds extra high, none, and auto properly into the NanoGPT backend. Previously it sent nothing at all and just picked a random effort.

### Others
Additionally, this also resolved other possible third-party extensions having problems with rendering buttons on the top bar in the future. If a button, similar to ST-Copilot had its own special CSS rules, SillyBunny didn't consider it a button with actual text, so it showed rendering issues with only an icon and text escaping the button bounds. Now it should grow to fit for similar CSS.

#### Testing
Tested both personally on my machine, any other errors I found (such as the search in CharacterLibrary) are bugged in the extension upstream due to my typically 4:3 browser size. ST-Copilot intercepts outgoing requests so it changes 'maximum' to 'high'. Any errors in that regard are on the extension upstream.

### DIFF
--- a/public/css/mobile-styles.css
+++ b/public/css/mobile-styles.css
@@ -1013,6 +1013,7 @@
         min-height: var(--sb-mobile-touch-target, 44px);
     }
 
+    /* SillyBunny divergence: text-labelled icon controls may expand beyond upstream square sizing. */
     .menu_button:is(.fa, .fa-solid, .fa-regular, .fa-brands):empty,
     .menu_button_icon:is(.fa, .fa-solid, .fa-regular, .fa-brands):empty,
     .right_menu_button:is(.fa, .fa-solid, .fa-regular, .fa-brands):empty,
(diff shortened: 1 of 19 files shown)


Geechan:
Added SB divergence comments and simplified the code a bit with commit f63d449

platberlitz:
> Added SB divergence comments and simplified the code a bit with commit [f63d449](https://github.com/SillyBunnyTeam/SillyBunny/commit/f63d449225113fbd16b2f39d1b62e6bdca997944)

Just fixed a minor mobile UI issue. Can review again after if needed

Geechan:
Should be OK to merge.

PR #731, opened 2026-08-04
fix(characters): align favourites bar across tabs

Geechan:
An earlier regression broke the spacing of the favourites bar between the different sub-tabs in the Characters menu. This PR fixes that by standardising the alignment and padding while removing the clipping caused by a prior patch. This also fixes another regression where the "Create New Chat Group" had its text hidden, inconsistent with the "Create New Character" button.

Should be checked on mobile, but works well on desktop.


platberlitz (approved):
Works on mobile, just needed to fix the test.

PR #741, opened 2026-08-06
feat: SillyBunny Debugger core extension

platberlitz:
iOS and Android cannot properly check console logs due to the nature of how they work. Probably some Android browsers can but definitely not iOS.
This extension is off by default and emulates the regular Browser Dev Tools to help diagnose mobile issues better and more quickly.
Also generates a diagnostic report that can be sent to us.
Privacy-focused so it redacts user credentials, and URL credentials. Uploads nothing and clears captured data when disabled.

Also good for MacOS WebKit layout inspection for problems with the UI.


TheLonelyDevil9:
Just working to commit some secret/privacy-related changes and we're gucci

TheLonelyDevil9:
It's still going since 6.5 hours

TheLonelyDevil9:
Summary of the changes in my commit:

-The debugger should no longer accidentally include private chat text, passwords, tokens, request contents, or detailed URLs in reports.
-When uncertain whether information is private, it hides it instead of risking exposure. `Fail-closed` approach.
-Turning the debugger off now properly removes its background listeners and temporary data, even if something fails during cleanup.
    -It will not overwrite another debugger already running on the page.
-The debugger button now works better with keyboards, phone notches, screen zooming, and mobile keyboards.
-Automated tests passed in Chromium and WebKit, and every GitHub check passed.

TheLonelyDevil9 (approved):
Good to go

Geechan:
Doing one last sanity check first.

Geechan:
Latest commit fixes some regressions found in the code, and a Linux incompatibility.

Details:

- Installed local/global debugger versions now take precedence over the bundled opt-in copy, preserving customized or newer installations and their enabled state.
- Removed the Linux-failing path-traversal test because it tested only Playwright fixture routing, not production behavior.
- Diagnostic reports now normalize extension IDs for case and third-party/ aliases and exclude the shadowed debugger alias.
- Added regression coverage for implementation precedence, manifest lookup, activation, default-off fallback, and normalized report state.

With this, I think it's good to merge.

Geechan (approved):

PR #743, opened 2026-08-06
fix: conversation mode send/save/cancel buttons non-functional

Geechan:
Attempts to address an issue where clicking the send button in Conversation Mode after deleting a prior message yields no results, and an issue where editing the message prevents the save and cancel buttons from working.

### DIFF
--- a/public/css/sillybunny-conversation.css
+++ b/public/css/sillybunny-conversation.css
@@ -1200,10 +1200,50 @@
     justify-content: flex-end;
 }
 
-.sb-conversation-message-edit-save,
+.sb-conversation-message-edit-control.menu_button {
+    --sb-conversation-edit-action-color: var(--color-primary, var(--sb-accent));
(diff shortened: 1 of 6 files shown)


Geechan:
Made some further tweaks to the edit functionality after discovering some problems.

- The quick access control toolbar no longer appears above edit message boxes, preventing a problematic layering effect.
- There can only be one edit box open at a time, reducing clutter and distractions.
- Save/cancel buttons replaced with icons found from Roleplay mode, for better learned consistency. Buttons are colour-coded the same.
- Save/cancel buttons now function as they should even when the message contents has not been edited. Save will always close the box while keeping edits, while cancel will always close the box while removing edits.
- Added a delete button in the edit menu as a backup contingency.

platberlitz (approved):
Works as intended for me on mobile, safe to merge.

PR #744, opened 2026-08-07
feat: add Terminal UI to Bundled Extras

Geechan:
Simply adds @platberlitz's official Terminal UI extension to our Official Extensions under the Home Screen Bundled Extras, placed next to the Moonlit Echoes theme tab.


Geechan:
Since this is an extremely minor change with very minimal code (could be considered a chore), I'll go ahead and merge this now.

PR #745, opened 2026-08-08
fix: allow bulk actions for group chats

platberlitz:
An upstream limitation where the bulk actions button is unavailable and can't be used. It can now be used.


Geechan:
Just added a small fix, letting group cards receive proper selected styling, and some more CSS contract coverage. Otherwise, merging.

PR #748, opened 2026-08-08
fix: Pathfinder bug fixes + making two features that weren't tied to anything real

platberlitz:
### Tool calling
- Claude didn't accept the way tool calling was formatted in this extension so it's now reformatted and fixed.
- The extension could pick the wrong lorebook if the AI named a book it couldn't write to. Refuses instead of trying to guess/fallback to something else completely.
- Merging an entry with itself permanently deleted an entry, now that is blocked.
- Now shows error when moving an entry to a waypoint that doesn't exist.
- Even when it saved nothing, duplicate detection wrote "entry was still saved". It now points the AI to the existing entry.
- The AI sent "false" as text for "permanently delete?" and it was read as "yes, delete". Now fixed.
- Queues two tools that write at the same time so they don't overwrite each other.

### General bug fixes
- When lorebooks are read-only, auto-summary no longer nags about it constantly. Message counts for Auto-summary no longer carry over between chats.
- Fixes summary tracking breaking for the very first entry in a book (ID 0).
- Switching Pathfinder agents no longer inherits the old agent's books/permissions.
- /pf-search actually searches by name now; /pf-remember respects write permissions.
- Fewer redundant lorebook loads per message (was up to thousands), quieter memory leaks, and a pile of dead code deleted.

## Features now actually having code instead of doing nothing
- Edits to pipeline prompts finally save.
- Standalone editor prompt changes finally save.
- The older "sidecar retrieval mode" now injects actual entries.
- The Pathfinder's map of the lorebook was never refreshed. You needed to restart to see the edits, now it refreshes on edit.
- The retrieval timeout setting didn't work, now it actually cancels slow retrievals.
- **Ask Before Executing**: per-tool checkboxes; checked tools show an Allow/Deny popup with a preview before changing your lorebook. Deny tells the AI to not do anything. If the popup can't appear, the call is denied.
- **Require tool use on every response**: the toggle now genuinely forces the AI to call at least one tool at the start of each reply (correct format per provider), but the follow-up reply itself is never forced, so responses can't get stuck in tool loops.


Geechan:
I did an audit of this PR and made a few changes:

- Pathfinder prompt failed to isolate when switching agents. This is fixed now
- Pathfinder_Remember now has a wrong-book fallback option for redundancy.
- [Split off the features from this PR into its own PR](https://github.com/SillyBunnyTeam/SillyBunny/pull/749), as they were getting out of scope for what is supposed to be a purely `fix` label.

Geechan (approved):

PR #754, opened 2026-08-11
fix: hardens save path for remote servers

platberlitz:
I usually use a remote VM to access SillyBunny.

The problem is when the connection drops mid-save. My phone sends the chat, the server saves it to disk and generates a new integrity token, but the connection dies before that token makes it back to my phone. So the file on disk has moved on and my phone hasn't, and there's no way for it to catch up, because that reply was the only place the new token existed.

From then on every save gets rejected with Chat Integrity Failed, and it keeps happening until you reload the page.

This makes it so if the phone's copy has every message that's already on disk plus additional ones, the server accepts it instead of rejecting it outright, a save that only adds messages can't lose anything. If an existing message was edited or removed, it still gets rejected.

Also prevents Chat Integrity Failed check loops where you enter a chat, click OK, refresh, then enter the chat, does the same thing, again and again... making you unable to use that chat.

Fixes the browser sending an empty chat when a pending metadata save fires in the middle of reopening a chat. The save's checks all pass because the character, chat and group haven't changed, but the messages have been cleared for reloading, so it tries to write a chat with no messages over the real one.

### DIFF
--- a/docs/upstream-touch-ledger.md
+++ b/docs/upstream-touch-ledger.md
@@ -50,17 +50,17 @@ This ledger tracks intentional SillyBunny divergence in upstream-origin files. I
 | Last reviewed | 2026-06-06 in-chat agent stop-button wiring. |
 | Owner | Refactor integrator. |
 
-### `src/endpoints/chats.js`, `public/script.js`, and `public/scripts/group-chats.js` - chat save integrity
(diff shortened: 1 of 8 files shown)


Geechan (approved):
Latest commit fixes an issue related to metadata saving problems, and adds an entry in `upstream-touch-ledger.md` as is required.

Should be good to merge now.

PR #759, opened 2026-08-14
chore: release E2E must now be manual

platberlitz:
Makes Chromium E2E test manual.


platberlitz:
Meh, it wasn't a main blocker so I'll keep as is.

PR #781, opened 2026-08-20
fix(generation): allow upstream responses to continue and finish

platberlitz:
Since iOS is insane, whenever I switch apps or even tabs for a couple minutes the connection to my remote server ends and it just cancels every response. Which means I have to micromanage it every single time.

## Added
- Generation requests carry an `X-Generation-Id`; the server keeps the reply and only stops on an explicit cancel.
- The page reattaches from its last byte via `POST /api/resumable-generations/resume`.
- Stop and page unload still cancel for real, a replayed POST can't leave two generations running.
- Finished replies are kept 15 minutes, an unused generation is dropped after an hour.

### DIFF
--- a/public/script.js
+++ b/public/script.js
@@ -330,6 +330,7 @@ import { compressRequest, setRequestCompressionConfig } from './scripts/request-
 import { canJumpToSwipeForMessage, canOpenSwipePickerForMessage, initSwipePicker } from './scripts/swipe-picker.js';
 import { bindIOSFastTapSendButton, isIOSWebKitPlatform } from './scripts/mobile-send-button.js';
 import { formatMobileStreamingPreview, getMobileStreamingBottomPinBehavior, getStreamingUpdateInterval, isAndroidStreamingPlatform, shouldReduceStreamingDomWork, shouldUsePlainTextStreamingPreview } from './scripts/mobile-streaming.js';
+import { fetchResumable } from './scripts/resumable-generation.js';
 import {
     CHAT_RENDER_LIFECYCLE_ROLLOUT_KEY,
     CHAT_RENDER_LIFECYCLE_ROUTE,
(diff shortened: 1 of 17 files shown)


TheLonelyDevil9:
Checking

TheLonelyDevil9:
Fixed some findings listed below (LLM comments for reference):

1. **One user could cancel another user's in-progress reply.** When the server's slots for tracked replies were full, a new request would kick out the oldest one — even if it belonged to someone else and was still being written. Everyone now gets their own small limit, and if the server is truly full, a new request simply loses the "resume after disconnect" perk instead of killing someone else's work.

2. **The server could hold too much in memory.** Each saved reply could take up to 32 MB with no overall cap, so many replies together could add up to gigabytes. (TLD - nearly 7 gigs)
There's now a shared 256 MB ceiling, and finished replies are cleared out first.

3. **Reconnect attempts could pile up.** A single reply could collect unlimited "continue from where I left off" connections and bog the server down. Replies now allow at most 8 at once, abandoned ones are cleaned up immediately, and frozen ones get cut loose instead of being fed data nobody is reading.

4. **A broken AI connection could leave a reply hanging.** If the Ollama backend's stream failed mid-reply, the reply could sit stuck for up to an hour. It now ends cleanly right away.

5. **Horde replies weren't covered.** Horde generations skipped the survive-a-disconnect system entirely. They're included now, and stopping them works properly too.

6. **Closing the tab on iPhone Safari didn't stop the AI.** iPhones don't send the usual "page is closing" signal, so the model kept running for up to an hour. The fix also listens for the signal iPhones do send, so closing the tab stops the generation — while just switching apps or locking the screen keeps your reply safe.

7. **Small cleanup:** re-requesting an already-finished reply now frees its stored copy immediately instead of leaving it in memory until cleanup.

8. **Tested:** 12 new automated tests covering each fix above (all focused and regression tests pass), code style checks pass, and a real server run confirmed replies resume with their original status and cancel on demand.

TheLonelyDevil9 (approved):
Good to go. Merging

PR #802, opened 2026-08-25
fix: High Resolution Blurriness

TheLonelyDevil9:
Does what it says on the tin. Reported by Kenzo in Marinara's Kitchen for 5120x2160. Fixes blurriness at high resolution while stacking on top of the GPU work done in PR #796 (fix: prevent the frontend from using so many resources to reduce GPU usage)

### DIFF
--- a/public/style.css
+++ b/public/style.css
@@ -202,16 +202,8 @@
 }
 
 html {
-    /*fix for chrome flickering on blurred divs*/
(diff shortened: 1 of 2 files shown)


TheLonelyDevil9:
Report from Kenzo confirms all is well, except one low-level blur conflict but that's inconsequential

PR #803, opened 2026-08-26
feat(ica): lorebook scout now directly sends to lorebook

platberlitz:
Before, Lorebook Scout was display only - now it actually can send to the lorebook if any keys are created.

Also has a dialogue box that confirms if you want to attach the Lorebook Scout as an auxiliary lorebook to the current chat/card.

### DIFF
--- a/docs/in-chat-agents-glossary.md
+++ b/docs/in-chat-agents-glossary.md
@@ -277,6 +277,8 @@ The token information pills describe the Companion request itself. Input tokens 
 
 ### Special Panel Controls
 
+Lorebook Scout - Creates a lorebook keyword after manually running it, then has an option to send it directly to the lorebook attached to the chat, or creates a Lorebook Scout that it places the entries to.
+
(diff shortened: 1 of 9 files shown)


TheLonelyDevil9:
Reviewing

TheLonelyDevil9:
- Kept imported lorebooks intact: SillyBunny held both a working copy and a copy embedded in the character card. The PR updated only the working copy, so saved cards could contain a blank or missing Scout entry. Both copies are now updated before saving.
- Required real trigger words: Scout output must now include 2–5 generated keywords. It no longer silently substitutes the entry title when keywords are missing.
- Made confirmation meaningful: when a chat has no lorebook, Scout saves its entry into the fallback book first, then asks before attaching that book. Declining leaves the book available but does not attach it to the chat/card.
- Made failed saves retryable: if saving fails, the next attempt retries the unsaved data instead of treating the cached entry as a duplicate and never saving it.
-Added regression tests for each case.

Popup’s wider character scope/panel button design untouched.

Merging.

Issue #706, opened 2026-07-31
[BUG] Re-saving Chats Upon Opening Them


TRUEDarkAngel:
### Environment

🪟 Windows

### System

Windows

### Version

Most recent staging

### Describe the problem

I accidentally started this issue on a different repo. Originally, the issue was that, when opening a chat, SB would remake the entire character card and chat, wiping all of the metadata and timestamps in the process. The issue was partially fixed via work done on a different extension because I thought it was related to it. It and the commit are linked below just in case.

https://github.com/platberlitz/sillytavern-character-colors/issues/4
https://github.com/SillyBunnyTeam/SillyBunny/commit/b81e06da9c59bad684403fec2453d2ad7869715a

This led to the issue with the character card being fixed, however the chat file is still re-created each time. It would be wonderful to get this fixed. Thank you!

### Please tick the boxes

- [x] I have explained the issue clearly, and I included all relevant info
- [x] I have checked that this [issue hasn't already been raised](https://github.com/platberlitz/SillyBunny/issues?q=is%3Aissue)
- [x] I have checked the [docs](https://docs.sillytavern.app/) ![important](https://img.shields.io/badge/Important!-F6094E)
- [x] I confirm that my issue is not related to third-party content, unofficial extension or patch. If in doubt, check with a new [user account](https://docs.sillytavern.app/administration/multi-user/) and with extensions disabled

TRUEDarkAngel:
Adding on to this just in general: in general, SB remakes the entire card when it is modified instead of just updating it. This leads to the same loss of metadata that this is trying to fix.

platberlitz:
> Adding on to this just in general: in general, SB remakes the entire card when it is modified instead of just updating it. This leads to the same loss of metadata that this is trying to fix.

Please open a separate issue for this one,

TRUEDarkAngel:
Alright. Moved that to its own thing.

platberlitz:
Try moving to the branch `fix/chat-resave-706` and see if that fixes the chat files specifically.

TRUEDarkAngel:
Sorry I had to walk away. Forgive me, but I'm not sure how to properly change to that branch. Could you maybe put it on staging?

TRUEDarkAngel:
Should I just download it manually? Sorry for the trouble.

platberlitz:
No worries. Go to the Customize -> Server then switch to the branch, or go use `git checkout fix/chat-resave-706`,

[...]

PR #818, opened 2026-09-03
fix(chats): stop rewriting chat files on open when extensions add styling metadata

cspiritsong:
### What this PR achieves
If you open a chat while using an extension (like Dialogue Colors or in-chat agents), SillyBunny secretly rewrites your chat file on disk and updates the file's modification timestamp — even if you literally just opened it and didn't touch a single key. On Windows, this resets your file dates and creates extra backup copies every time you click a chat.

This happens because extensions add temporary visual styling or token counts to messages in memory when they render. SillyBunny compares the in-memory chat against what's on disk, sees these new extension tags, mistakes them for a user edit, and saves over the file.

This PR stops those unnecessary file rewrites while keeping your file timestamps untouched.

---

### How it achieves its goal
In `src/endpoints/chats.js`, we told the server's change-detection check (`normalizeChatMessageExtraForComparison`) to ignore a small list of known temporary extension tags (`dialogue_colors`, `character_colors`, `display_text`, `token_count`, `qr_rendered`).

If the *only* difference between the open chat and the file on disk is one of these styling tags, the server skips the disk write and leaves the original file alone.

This check is strictly for deciding whether to save on open: if you actually type a message or edit text, it still saves everything (including all extension data) in full.

---

### Steps to test (using the test card & chat)

Test files:
- `Inspector-Bun.png` (Character card)
- `Inspector-Bun-Case001.jsonl` (Test chat containing extension styling tags)

**In current staging (the bug):**
1. Put the card and chat into your SillyBunny install. Note the chat file's modification date on your computer.
2. Open SillyBunny and open Inspector Bun's chat.
3. Look at your server terminal: you will see `[Backup] chat-save-written` fire.
4. Check the file on disk: the file timestamp changed to right now, even though you did nothing.

**With this patch (the fix):**
1. Do the exact same steps.
2. The server terminal will now show `[Backup] chat-save-skipped (reason: unchanged)`.
3. Check the file on disk: the timestamp and file contents remain completely untouched.
4. Now actually edit a word in the chat and hit save: it saves immediately, updating the file and keeping all extension data.
5. Automated tests: `npm run test:unit -- chat-integrity-rotation.test.js` (44/44 pass).

### DIFF
--- a/src/endpoints/chats.js
+++ b/src/endpoints/chats.js
@@ -152,7 +152,33 @@ function normalizeChatMessageExtraForComparison(extra) {
         return extra;
     }
 
-    const normalized = JSON.parse(JSON.stringify(extra));
+    const normalize = (value, seen) => {
+        if (value === null || typeof value !== 'object') {
+            return value;
+        }
+
+        if (seen.has(value)) {
+            return value;
+        }
+
+        seen.add(value);
+
+        if (Array.isArray(value)) {
+            const result = value.map((item) => normalize(item, seen));
+            seen.delete(value);
+            return result;
+        }
+
+        const result = {};
+        for (const key of Object.keys(value).sort()) {
+            result[key] = normalize(value[key], seen);
+        }
+
+        seen.delete(value);
+        return result;
+    };
+
+    const normalized = normalize(JSON.parse(JSON.stringify(extra)), new WeakSet());
     if (Object.hasOwn(normalized, 'file')) {
         normalized.files = Array.isArray(normalized.files) ? normalized.files : [];
         if (normalized.file) {
--- a/tests/chat-integrity-rotation.test.js
+++ b/tests/chat-integrity-rotation.test.js
@@ -335,6 +335,51 @@ describe('chat integrity rotation', () => {
         expect(backupFiles.filter(fileName => fileName.startsWith('chat_pre_write_test_card_'))).toHaveLength(0);
     });
 
+    test('persists extension extra changes and rotates integrity when metadata changes without text changes (#706)', async () => {
+        const { trySaveChat } = await import('../src/endpoints/chats.js');
+        const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'sillybunny-chat-extra-persistence-'));
+        const chatFile = path.join(tempDir, 'chat.jsonl');
+        const backupDir = path.join(tempDir, 'backups');
+        await fs.mkdir(backupDir);
+
+        const baselineChat = [
+            { user_name: 'User', character_name: 'Test Card', chat_metadata: { integrity: 'initial-extra-slug' } },
+            { name: 'Test Card', is_user: false, mes: 'Message text.', send_date: 'August 1, 2026', extra: { selected_media_index: 0 } },
+        ];
+        await fs.writeFile(chatFile, baselineChat.map(row => JSON.stringify(row)).join('\n') + '\n', 'utf8');
+
+        // User selects a different media item in a gallery or updates styling metadata without editing text:
+        const incomingPayload = [
+            { user_name: 'User', character_name: 'Test Card', chat_metadata: { integrity: 'initial-extra-slug' } },
+            {
+                name: 'Test Card',
+                is_user: false,
+                mes: 'Message text.',
+                send_date: 'August 1, 2026',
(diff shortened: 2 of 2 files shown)


cspiritsong:
### Test Chat Fixture (`Inspector-Bun-Case001.jsonl`)

To easily test without setting up third-party extensions, create a chat file with this exact content (which simulates an extension injecting `dialogue_colors` and `token_count` into `message.extra`):

```json
{"user_name":"User","character_name":"Inspector Bun","create_date":"August 1, 2026 12:00pm","chat_metadata":{"integrity":"f2cd5e11-f9c0-4cf1-976e-a817d6d6d74d"}}
{"name":"Inspector Bun","is_user":false,"is_system":false,"send_date":"August 1, 2026 12:00pm","mes":"The Burrow Precinct smells faintly of tea and wet wool. Case File 001 is open.","extra":{}}
{"name":"User","is_user":true,"is_system":false,"send_date":"August 1, 2026 12:01pm","mes":"Checking in with the calibration desk.","extra":{}}
{"name":"Inspector Bun","is_user":false,"is_system":false,"send_date":"August 1, 2026 12:02pm","mes":"Docket confirmed. Ready for testing.","extra":{"dialogue_colors":{"0":"#3498db"},"token_count":38}}
```

Place it under `data/default-user/chats/default_Seraphina/Inspector-Bun-Case001.jsonl` (or any character), set its file timestamp to a past date, and open it in SillyBunny.

cspiritsong:
### Architecture & Implementation Update (Rethink 1)

Following an internal code audit on metadata persistence, we updated this branch to remove comparison-time key stripping in `src/endpoints/chats.js`.

**Why comparison-time stripping was reverted:**
Stripping `message.extra` keys (such as `dialogue_colors`, `token_count`, `selected_media_index`, or `display_text`) during backend equality comparison introduces a silent data-loss risk: if a user customizes dialogue colors or selects a gallery media index without modifying message text, the server would consider the payload identical to disk, skip writing, and revert the user's styling/selection upon reload.

**What the updated patch does:**
- Preserves all arbitrary `message.extra` fields as strictly persistence-significant on the server.
- Adds an automated regression test in `tests/chat-integrity-rotation.test.js` verifying that metadata changes to `message.extra` (specifically `selected_media_index` and `dialogue_colors`) rotate integrity and commit cleanly to disk without requiring message text edits.
- All 43 tests pass in `tests/chat-integrity-rotation.test.js`, and ESLint is clean.

cspiritsong:

We incorporated GPT-5.6-Luna's canonical key-sorting preservation model for `normalizeChatMessageExtraForComparison` in `src/endpoints/chats.js`:
- All extension-owned metadata keys (`selected_media_index`, `dialogue_colors`, `token_count`, etc.) are preserved as comparison-significant.
- Applied recursive deterministic key sorting to object metadata so comparisons remain completely stable without mutating arrays or stripping extension state.
- Verified on Windows Canary: 43/43 unit tests pass cleanly, and ESLint is clean. (Commit: `d4fb09ba2`)

cspiritsong:
Closing this, and I want to be direct about why: this patch does nothing. It's a no-op and it should not be merged.

Here's what I found going back over it:

1. My own test passes without my fix. I reverted my change to `src/endpoints/chats.js`, kept the new test, and it still passed — 43/43. The test I wrote to prove the fix never proved anything.

2. The mechanism is redundant. My patch recursively sorts object keys in `normalizeChatMessageExtraForComparison` so key order can't produce a false difference. But the comparison goes through `isDeepStrictEqual`, which already ignores key order. I was defending against something that can't happen.

3. I checked it directly. I ran the current staging normalizer and my patched one over the same inputs — different key order, nested key order, arrays of objects, genuinely different values, added keys. Same verdict every time. No behavioural difference in any case I could construct.

So it's 28 lines of dead code with a test that proves nothing. Not worth your review time.

On #706 itself: I couldn't reproduce it on a clean 1.7.0 install either — no chat file changes across idle time or repeated reloads. I don't have a reproduction, which is probably how I ended up writing a fix that fixes nothing. If I get a real repro I'll come back with an actual failing test first.

Also: I force-pushed this branch after it was out of draft, which your PR template asks contributors not to do. My mistake — I'll keep branches in draft until they're final.

Issue #340, opened 2026-06-04
[Bug] Chats Seemingly Cloning


TRUEDarkAngel:
Sometimes, for unknown reasons, the Recent Chats will show the same chat at a different point in its history, as seen in the image. While it's generally harmless, it's annoying and makes no sense. It may have something to do with previous issues, however I have no idea.

<img width="1608" height="417" alt="Image" src="https://github.com/user-attachments/assets/c29b3370-580d-47ba-bfd4-2ec5b5f83990" />

TRUEDarkAngel:
It does create new files in the folder too.

platberlitz:
Tried to fix with #343 - check if it works?

TRUEDarkAngel:
After using it for a bit last night and checking back today, it seems not.

platberlitz:
#349 was pushed - see if that works/helps?

TRUEDarkAngel:
Seems to be better...will test a bit longer and let you know.

TRUEDarkAngel:
After playing around for long enough, swipes seem to bug out less but still sometimes do.

(Real quick just while I'm here, what's the best model to use for the Prose Polisher outta curiosity.)

platberlitz:
> After playing around for long enough, swipes seem to bug out less but still sometimes do.

Let me see if I can harden that.

> (Real quick just while I'm here, what's the best model to use for the Prose Polisher outta curiosity.)

I use Gemma 4 31B myself.

platberlitz:
#350 was just pushed, please check when you have time.

TRUEDarkAngel:
The previous fix did seem to mostly work so I'll try to make this work tomorrow, I'll lyk if I find anything.

TRUEDarkAngel:
Seems to still exist at times. may be related to agents, unknown.

platberlitz:
@TheLonelyDevil9, maybe you can investigate when you have time.

TRUEDarkAngel:
It may be related to an extension I have or an agent but I really have no clue.

TRUEDarkAngel:
(Small thing while I'm here, what would you people suggest using for memory or summarization or retrieval or whatever in SB? An extension or a feature or what? Just curious given how much this thing changes ST.)

platberlitz:
> (Small thing while I'm here, what would you people suggest using for memory or summarization or retrieval or whatever in SB? An extension or a feature or what? Just curious given how much this thing changes ST.)

[...]

PR #819, opened 2026-09-03
fix(chats): invalidate queued saves across swipe and lifecycle changes to prevent chat cloning (#340)

cspiritsong:
### What this PR achieves (in plain English)

If you swipe through alternative replies quickly, run in-chat agents (like Prose Polisher or companions), or switch chats while a message is processing, SillyBunny could sometimes take an older slice of your chat history and save it into a brand new `.jsonl` file on your computer.

When this happened, your "Recent Chats" list would suddenly show two (or more) copies of the exact same chat, with one copy stuck at an older message count. 

This PR ensures that whenever you swipe, switch chats, or start a new turn, any pending or queued saves from an earlier swipe or turn are immediately marked as outdated and cancelled. Older chat snapshots will never be written over newer ones or saved as duplicate files.

---

### Why it happened (the technical root cause)

1. **`chatSaveQueue` Dequeue Blindness:** While commit `b77b1fbcc` snapshotted the chat data and filename at enqueue time, the active chat generation counter (`chatGeneration`) was not snapshotted into the queued arguments.
2. **Missing Dequeue Guard in `saveChatImmediately`:** `getDebouncedChatSaveAbortReason` only guarded the initial debounce timer. Once a save transitioned onto the serialized `chatSaveQueue` promise chain, `saveChatImmediately` ran without verifying whether the active chat generation or entity identity had changed while it was waiting in the queue.
3. **Swipes Did Not Invalidate Prior Saves:** `standardSwipe()` did not increment `chatGeneration`. As a result, an in-flight save or a debounced agent post-processing pass from swipe $N-1$ still matched `scheduledGeneration === currentGeneration` when swipe $N$ was active. When that older save finally ran, it wrote an outdated message slice, or if navigation occurred, created a duplicate `.jsonl` file on disk.

---

### How it achieves its goal

1. **Queued Abort Guard:** Added `getQueuedChatSaveAbortReason` in `public/scripts/chat-save-guard.js` to validate that queued saves still match the active character, group, chat file, and lifecycle generation.
2. **Save Queue Snapshotting:** In `saveChat` (`public/script.js`) and `saveGroupChat` (`public/scripts/group-chats.js`), snapshotted `scheduledGeneration: chatGeneration`, `scheduledCharacterId`, `scheduledGroupId`, and `scheduledChatId` into the queued payload.
3. **Immediate Dequeue Check:** `saveChatImmediately` and `saveGroupChatImmediately` now check `getQueuedChatSaveAbortReason` immediately before executing and cleanly abort if generation or context changed while waiting in the queue.
4. **Swipe Invalidation:** `standardSwipe` now calls `incrementChatGeneration()` whenever the swipe ID changes, instantly invalidating any pending debounced saves or queued saves from prior swipes.

---

### How we proved it works (reproduction & verification)

Tested across both Linux and native Windows 11 (`SillyBunny-Canary` on Bun/Node):

1. **Before the fix (in current staging):**
   - Scheduling a save and immediately advancing the chat generation (via navigation or swipe) leaves the queued save unaware of the generation change; when the queue drains, `saveChatImmediately` executes the stale save anyway.
   - Rapid swiping leaves debounced agent post-processing passes with an active generation match, allowing older swipe snapshots to save after a new swipe is rendered.
2. **With this patch (on Canary):**
   - The queued abort guard detects the generation change at dequeue and logs:
     `saveChatImmediately aborted, but chat generation changed while queued.`
   - Swiping increments `chatGeneration`, cancelling older pending saves before they can land.
3. **Automated Unit Tests:**
   - Run on Windows Canary: `npm run test:unit -- chat-save-guard.test.js chat-integrity-rotation.test.js character-chat-resolver.test.js`
     **Result:** 60 passed, 60 total (all green).
   - Full chat suite (8 test files covering rotation, recovery, rename, and destructive write protection):
     **Result:** 122 passed, 122 total (all green).
4. **Clean Mergeability:**
   - Clean single commit (`753503051`), rebased directly on `staging` HEAD. Zero merge conflicts, zero rebase needed ("clean as heck, just merge it").
   - `npm run lint` is 100% clean (0 errors, 0 warnings).

### DIFF
--- a/public/script.js
+++ b/public/script.js
@@ -39,7 +39,7 @@ import {
 } from './scripts/textgen-settings.js';
 import { shouldRestoreTextGenStatusOnStartup } from './scripts/textgen-startup-status.js';
 import { normalizeCharacterChatName, resolveCharacterChatNameForLoad } from './scripts/character-chat-resolver.js';
-import { getDebouncedChatSaveAbortReason } from './scripts/chat-save-guard.js';
+import { getDebouncedChatSaveAbortReason, getQueuedChatSaveAbortReason } from './scripts/chat-save-guard.js';
 import { getCharacterDefinitionFormValues, getSuspiciousEmptyCharacterDefinitionSave } from './scripts/character-save-guard.js';
 // SillyBunny: keep model-produced chat filenames behind a strict, independently tested parser.
 import { CHAT_LABEL_TITLE_LIMIT, extractGeneratedChatLabel, normalizeGeneratedChatLabel, truncateChatLabelText } from './scripts/chat-label.js';
@@ -10678,6 +10678,7 @@ export function saveChat(...saveChatArguments) {
         const chatData = cloneChatSavePayload(sourceChatData);
         const metadataSnapshot = structuredClone({ ...chat_metadata, ...(options.withMetadata || {}) });
         const activeCharacter = characters[this_chid];
+        const currentGeneration = chatGeneration;
         queuedSaveArguments = [{
             ...options,
             chatData,
@@ -10686,6 +10687,10 @@ export function saveChat(...saveChatArguments) {
             characterName: activeCharacter?.name,
             avatarUrl: activeCharacter?.avatar,
             wasGroupChat: Boolean(selected_group),
+            scheduledGeneration: currentGeneration,
+            scheduledCharacterId: this_chid,
+            scheduledGroupId: selected_group,
+            scheduledChatId: getCurrentChatId(),
         }];
     }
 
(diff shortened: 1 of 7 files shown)


cspiritsong:
### Verification & How to Test in 30 Seconds

To verify this without having to manually race rapid swipes or LLM streaming latency, we added a dedicated test suite directly into `tests/chat-cloning-lifecycle.test.js`.

You can run it right from the `tests/` directory:

```bash
cd tests && npm run test:unit -- chat-cloning-lifecycle.test.js chat-save-guard.test.js
```

### What the test proves:
1. **The Race Condition:** Simulates a queued save in `chatSaveQueue` when the user swipes or switches chats (advancing active generation from 1 to 2). Without our guard, the unshielded dequeue executes the stale save, writing outdated history or creating a new file.
2. **The Fix:** With `getQueuedChatSaveAbortReason`, the queued save detects that `scheduledGeneration (1) !== currentGeneration (2)`, cleanly aborts, and prevents any network call or disk write.
3. **Normal Saves Untouched:** Valid saves with matching generation and character/group context pass through and save normally.

Tested and verified on Windows 11 under Bun and Node.js. All 122 chat tests pass with zero regressions, and ESLint is clean.

cspiritsong:
### Architecture & Implementation Update (Rethink 2)

Following an internal code audit, we hardened `saveChatImmediately` and `getQueuedChatSaveAbortReason`:

**What was caught and fixed:**
1. **Positional argument normalization:** In `saveChatImmediately`, the queued save abortion check previously ran *before* legacy positional arguments were normalized. This caused any legacy or extension call using positional signatures to abort prematurely because `scheduledCharacterId` was undefined. Positional normalization has been moved to the very top of `saveChatImmediately`.
2. **Direct unqueued save passthrough:** `getQueuedChatSaveAbortReason` was tightened so identity and generation checks only evaluate when scheduled fields are defined (`!== undefined`). Unqueued direct saves are never blocked.
3. **Swipe state commitment:** Verified that `syncMesToSwipe` is called prior to swiping away, ensuring that active message text and extra metadata are safely committed to the swipe array in memory before `chatGeneration` advances.

**Verification:**
- Added unit tests in `tests/chat-save-guard.test.js` confirming unqueued direct saves pass through without abortion.
- All 20 lifecycle and guard tests pass across `tests/chat-save-guard.test.js` and `tests/chat-cloning-lifecycle.test.js`.
- ESLint is clean, and the branch is cleanly rebased on staging HEAD (`725871f3c`).

cspiritsong:

We refined `getQueuedChatSaveAbortReason` and `saveChatImmediately` per Luna 5.6's drift model:
- Positional arguments are normalized before abort checks to preserve backward compatibility for legacy callers.
- The abort guard evaluates drift whenever scheduled fields are defined, including catching departure drift when the current chat becomes undefined (closed or navigated away). Direct unqueued saves remain unaffected.
- Verified on Windows Canary: 21/21 unit tests pass cleanly, and ESLint is clean. (Commit: `a1056e223`)

cspiritsong:
Final Windows Canary verification for the current head (`a1056e223`):

- Confirmed the server before and after the run via `/version`: `fix/chat-cloning-340` at `a1056e22`, SillyBunny 1.7.1.
- Focused chat save/lifecycle coverage: 65/65 passed under Node 22.23.2 and Bun 1.4.0.
- Differential check: reverting the source change while retaining the regression tests causes four tests to fail and two suites not to load, confirming the patch is load-bearing.
- Full Playwright pack against the confirmed Canary: 589 passed, 3 skipped, 2 failed out of 594 in 18.9 minutes.

The two Playwright failures are outside this patch's surface: a screenshot marker-width threshold missed by 0.6 px, and a mobile composer geometry assertion. This PR changes only the chat save queue/lifecycle code and its tests; it has no screenshot, composer, or CSS changes. The mobile composer failure was also reproduced on the untouched stock baseline (`main` at `3021edca`). No Playwright failure occurred in the changed chat-save/lifecycle surface.

Issue #373, opened 2026-06-08
[Bug] Inane Amount of Backups + VERY Bloated Chat Logs


TRUEDarkAngel:
While yes, having backups is good, it becomes a real problem when SB creates a backup of a really long chat if you even OPEN it. I checked my `backups` folder and there are MULTIPLE BACKUPS happening every MINUTE of files that are over 30 MB, leading to multiple GB of bloat. It's also backing up the settings file way too often. While I am aware that this is base ST functionality, SB's logs seem oddly larger than ST logs for some reason (since one log continued from ST is 4x the size of the ST file despite only adding a good 30-40 messages to it) and even then THIS MANY BACKUPS is just a little absurd. It would be wonderful if something could be done.

<img width="617" height="340" alt="Image" src="https://github.com/user-attachments/assets/0032f449-055d-4114-a840-b74aeac59726" />

TRUEDarkAngel:
It would be far more manageable if the backups were at least compressed with something like 7-Zip/NanaZip, but between the massive and bloated file size, it is utterly insane. The reason for the log bloat from what I can see is that it retroactively includes EVERY AGENT REGEX in EVERY MESSAGE of the log which is just utter madness if you're like me and like to use agents to replace CSS and HTML with the regex HTML function. This is just utterly ridiculous and I hope it gets fixed sooner rather than later.

The included image is the result of me manually extracting messages from the chat files and placing them in individual text files.

<img width="614" height="68" alt="Image" src="https://github.com/user-attachments/assets/698f9d4b-05f6-46bf-8459-0b5c008e1015" />

TRUEDarkAngel:
Not to mention that, with this setup, it creates like SIX BACKUPS after a SINGLE swipe without even a post-response agent going off.

TRUEDarkAngel:
I saw the commit. It would be nice if you could retroactively apply the shortening to older chat files. Or at least have a way to convert. Also it still shoves like 8 backups out after a single swipe, like almost one new backup for each agent.

platberlitz:
I'll add that.

TRUEDarkAngel:
13.8 MB is certainly better than 33.3 MB. But you should probably add temporary logging for the backups since it's still making too many of them.

platberlitz:
Closing for now.

TRUEDarkAngel:
It does prewrite->normal->prewrite->normal->prewrite->normal for each swipe. There is one less pair on a new message.

[...]

PR #820, opened 2026-09-03
fix(chats): suppress redundant pre-write backups during deferred save runs to prevent ring eviction (#373)

cspiritsong:
### What broke?
When running in-chat agents (like Trackers, Prose Polisher, or Companions) or generating messages, SillyBunny performs multiple intermediate saves in the background before the turn completes. 

Although regular post-save backups are deferred until the turn finishes, SillyBunny was creating an unthrottled `chat_pre_write_` backup on every single intermediate save. For a chat with 3 or 4 agent steps, this generated **4 to 6 full-sized `.jsonl` files in `backups/` on every single message turn**.

Even worse: because the pre-write backup ring buffer only keeps the 3 most recent files, a 4-step agent turn would push out and delete the true pre-turn backup from before the message was generated, leaving only intermediate half-baked agent drafts in the recovery ring.

### Why did it break?
In `src/endpoints/chats.js` (`trySaveChatLocked`), `backupChatPreWrite()` was called unconditionally whenever incoming content differed from the on-disk file, completely ignoring the `deferBackup` option sent by the client.

### How is it fixed?
1. The server now tracks active deferred save sequences (`activeDeferredChatPreWrites`) by file path.
2. On the first write of a deferred run, `backupChatPreWrite()` captures the pristine pre-turn baseline on disk as the recovery anchor.
3. Subsequent intermediate saves within the same deferred sequence skip writing redundant pre-write snapshots to disk.
4. When the sequence finishes with its final non-deferred save, the deferred tracker is cleared, intermediate pre-write churn is skipped, and the standard post-save backup captures the final completed reply.
5. Standard, non-deferred chat saves continue to capture pre-write snapshots exactly as before.

### How was it tested?
- Tested on both Linux Node.js and Windows Canary (`C:\Users\badiy\SillyBunny-Canary`, port 4445).
- All 43 tests in `tests/chat-integrity-rotation.test.js` pass cleanly.
- Added a permanent regression test `preserves pre-turn baseline and avoids ring eviction across multi-pass agent sequences (#373)` that exercises 4 consecutive deferred agent passes followed by a final save and verifies that:
  1. Exactly 1 pre-write backup is created (preserving the pristine pre-turn baseline).
  2. Exactly 1 post-save backup is created (preserving the final reply).
  3. Intermediate disk-thrash snapshots are eliminated and zero eviction occurs.
- ESLint passed with 0 errors and 0 warnings.

### DIFF
--- a/public/script.js
+++ b/public/script.js
@@ -40,6 +40,7 @@ import {
 import { shouldRestoreTextGenStatusOnStartup } from './scripts/textgen-startup-status.js';
 import { normalizeCharacterChatName, resolveCharacterChatNameForLoad } from './scripts/character-chat-resolver.js';
 import { getDebouncedChatSaveAbortReason } from './scripts/chat-save-guard.js';
+import { getChatBackupSaveOptions } from './scripts/chat-backup-sequence.js';
 import { getCharacterDefinitionFormValues, getSuspiciousEmptyCharacterDefinitionSave } from './scripts/character-save-guard.js';
 // SillyBunny: keep model-produced chat filenames behind a strict, independently tested parser.
 import { CHAT_LABEL_TITLE_LIMIT, extractGeneratedChatLabel, normalizeGeneratedChatLabel, truncateChatLabelText } from './scripts/chat-label.js';
@@ -212,6 +213,7 @@ import {
     loadFileToDocument,
     getSanitizedFilename,
     getStringHash,
+    uuidv4,
 } from './scripts/utils.js';
 import {
     TOOLING_UI_HYDRATION_STATUS,
@@ -10665,7 +10667,7 @@ export async function flushPendingChatSaves({ silent = false } = {}) {
  */
 export function saveChat(...saveChatArguments) {
     const [firstArgument] = saveChatArguments;
-    const options = firstArgument && typeof firstArgument === 'object'
+    let options = firstArgument && typeof firstArgument === 'object'
         ? firstArgument
         : saveChatArguments.length === 0
             ? {}
@@ -10673,6 +10675,7 @@ export function saveChat(...saveChatArguments) {
     let queuedSaveArguments = saveChatArguments;
 
     if (options) {
+        options = getChatBackupSaveOptions(options, JSON.stringify([characters[this_chid]?.avatar, getCurrentChatId(), chatGeneration]), uuidv4);
         const mesId = options.mesId;
         const sourceChatData = Array.isArray(options.chatData)
             ? options.chatData
@@ -10703,7 +10706,7 @@ export function saveChat(...saveChatArguments) {
     return saveTask;
 }
 
-async function saveChatImmediately({ chatName, withMetadata, metadataSnapshot, mesId, force = false, chatData = undefined, throwOnError = false, deferBackup = false, allowShrink = false, activeChatName, characterName, avatarUrl, wasGroupChat = false } = {}) {
+async function saveChatImmediately({ chatName, withMetadata, metadataSnapshot, mesId, force = false, chatData = undefined, throwOnError = false, deferBackup = false, deferSequenceId, allowShrink = false, activeChatName, characterName, avatarUrl, wasGroupChat = false } = {}) {
     if (wasGroupChat || (selected_group && !activeChatName)) {
         toastr.error(t`Operation was aborted to prevent data corruption.`, t`saveChat called for a group chat`);
         throw new Error('saveChat called for a group chat');
@@ -10764,6 +10767,7 @@ async function saveChatImmediately({ chatName, withMetadata, metadataSnapshot, m
                 avatar_url: resolvedAvatarUrl,
                 force: force,
                 deferBackup: Boolean(deferBackup),
+                deferSequenceId,
                 allowShrink: Boolean(allowShrink),
             }),
         });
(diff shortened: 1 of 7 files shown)


cspiritsong:
### Verification & How to Test in 30 Seconds

To verify this without having to manually set up multi-step in-chat agents or wait for generation latency, we added a dedicated automated test directly into `tests/chat-integrity-rotation.test.js`.

You can run it right from the `tests/` directory:

```bash
cd tests && npm run test:unit -- chat-integrity-rotation.test.js
```

### What the test proves:
1. **The Flaw (#373):** Simulates a multi-pass agent turn (4 consecutive intermediate saves with `deferBackup: true` followed by 1 closing save). On unpatched code, 4 separate pre-write backup files were written to disk, pushing out and evicting the true pre-turn baseline from the 3-entry ring buffer (`PRE_WRITE_BACKUP_RING_SIZE = 3`).
2. **The Fix:** With sequence tracking in `trySaveChatLocked`, intermediate pre-write writes during deferred sequences are cleanly skipped. Exactly 1 pre-write snapshot is preserved (holding the pristine pre-turn baseline) and exactly 1 post-save backup is written upon completion. Zero ring eviction, zero disk thrash.
3. **Clean Baseline:** Standard non-deferred saves continue to capture pre-write snapshots normally.

Tested and verified on Windows 11 (Bun and Node.js) and Linux. All 43 tests pass cleanly and ESLint is 100% clean.

cspiritsong:
### Architecture & Implementation Update (Rethink 3)

Following an internal code audit, we hardened the deferred backup sequence tracking in `src/endpoints/chats.js`:

**What was caught and fixed:**
1. **Explicit sequence tokens (`deferSequenceId`):** In the initial implementation, active deferred tracking relied on a coarse 60-second window. If an agent sequence crashed or was cancelled mid-run without sending a closing save, a subsequent unrelated manual user edit could arrive within that window and be mistaken for the closing save, skipping its pre-write backup. We bound sequence suppression to an explicit `deferSequenceId`. Unrelated manual user saves (lacking a matching sequence ID) are never suppressed and always capture pre-write backups.
2. **Chat rename cleanup:** Added explicit cleanup in `/rename` (`activeDeferredChatPreWrites.delete(pathToOriginalFile)`) to prevent leaking original file paths in active tracking.
3. **Sequence closure isolation:** Closing saves only skip intermediate pre-write snapshots if their `deferSequenceId` matches the active sequence.

**Verification:**
- Added automated regression test `abandoned deferred sequence does not suppress pre-write backup for subsequent user edits (#373)`.
- All 44 tests pass in `tests/chat-integrity-rotation.test.js`.
- ESLint is clean, and the branch is cleanly rebased on staging HEAD (`949de8164`).

cspiritsong:

We adopted GPT-5.6-Luna's token-bound deferred pre-write backup state machine in `src/endpoints/chats.js`:
- Bounded deferred backup suppression strictly to normalized file paths and explicit `deferSequenceId` tokens.
- Guaranteed that unrelated manual user saves (lacking a matching sequence ID) always take pre-write backups and clear any stale abandoned sequence state.
- Cleaned up active sequence tracking on `/rename` and deletion endpoints to prevent path leaks.
- Verified on Windows Canary: 44/44 unit tests pass cleanly, and ESLint is clean. (Commit: `aa7bbf3b8`)

cspiritsong:
Moving this to draft. Going back over it with fresh eyes I found it isn't actually wired up yet, and I don't want it reviewed as if it works.

Two problems:

1. The client never sends the sequence token this depends on. The server logic keys off `deferSequenceId`, but none of the save paths in `public/` actually emit it — the agent helper and the solo/group queues only send `deferBackup`. So in a real agent run the suppression never activates; my earlier tests passed because they fed the token in directly. The mechanism works in isolation but isn't connected to anything end to end.

2. A failed pre-write backup is treated as success. The backup helper swallows errors, and the suppression state is committed anyway, so a transient disk/permission failure could leave a destructive write with no recoverable prior state while every request reports success.

Neither of these is the kind of thing I want a maintainer to find. I'm parking this and re-planning it properly — wire the token through the client, and make the backup requirement actually enforce a snapshot before it suppresses.

Separately: I had three PRs open at once all touching chat persistence (#818, #819, this one). That was a mistake — reviewing them side by side turned up things that don't show up in any of them individually. Landing them one at a time from now on: #818 is closed (it was a no-op — explained there), #819 first, this one after.

Same note on the post-draft force-push here. Won't happen again.

cspiritsong:
Correction to my previous comment: I said my Windows test machine was down. That was wrong — it was up, and I'd misdiagnosed a connection timeout on my end as the machine being offline. I've now run the Node/Bun parity checks I said I couldn't run.

Both pass: 44/44 on the chat-persistence suite under Node 22.23.2 and under Bun 1.4.0.

This doesn't change the outcome — the PR still isn't ready, for the reason in point 1 of my last comment: the client never sends the `deferSequenceId` token this depends on, so the suppression never activates in a real agent run. I've confirmed that on the Windows tree as well (no hits for `deferSequenceId` anywhere in `public/`). Staying in draft until I wire the token through properly and make the pre-write backup enforce a snapshot rather than assume one.

Apologies for the noise — I'd rather correct a wrong reason than leave it standing.

PR #824, opened 2026-09-05
fix: protect drafts and saved settings when using Story Mode

platberlitz:
I added the support Story Mode needed from SillyBunny itself. Previously, controlling one continuation could involve changing shared settings, and failed saves didn't reliably report failure. This prevents Story Mode from changing your setup and breaking things. Useful for the eventual future PR for Story Mode as a full feature implementation.

**What Changed**
- Replies can have their own length limit without changing your preset or accidentally limiting an agent's separate reply.
- Auto-continue can be skipped for one action without switching it off in your saved settings.
- Retry can leave the message box alone, instead of sending your unfinished draft or removing the reply it was meant to continue.
- Agents can be paused temporarily without changing their saved on/off switches. Saving an agent or changing chats won't accidentally make that pause permanent. Waiting agents and companions also check whether they're still allowed to run.
- Short rewrites can keep a thinking model's normal thinking allowance. Reaching a passage's length limit finishes the reply normally, rather than treating it as cancelled and skipping the usual follow-up work.
- Extensions can now confirm whether chat and character changes actually saved. Failed character saves no longer update the local card as though they succeeded.

These controls are optional. Existing settings keep their usual behaviour unless an extension asks to use them. Story Mode's layout, editing and Undo/Redo fixes are in its own repository, not this PR.


platberlitz:
Ideally: Merge this first before working on Story Mode.

OR we can draft this first and work on Hopper for now, which is simpler.

Up to you guys.

PR #831, opened 2026-09-07
chore: update nanogpt provider list

DeathStalker471:
In pull request #830 Nanogpt provider selection functionality was restored to SillyBunny inline with how it is handled in SillyTavern upstream. However, currently in upstream SillyTavern provider functionality is handled via a hard coded list in `public/scripts/textgen-models.js`, and is currently out of date with both upstream SillyTavern staging and the master Nanogpt provider list.

This PR updates the hard coded list inline with the current master list, this necessitates modifying the base SillyTavern file `public/scripts/textgen-models.js`.

Additionally this model list may have slight ordering difference from the upstream SillyTavern version. This version maintains the default order pulled from the master Nanogpt list meanwhile the SillyTavern version has moved some around to be in order. This makes no difference to the user interface as it is sorted automatically.

### DIFF
--- a/public/scripts/textgen-models.js
+++ b/public/scripts/textgen-models.js
@@ -119,6 +119,14 @@ const OPENROUTER_PROVIDERS = [
  */
 const NANOGPT_PROVIDERS = [
     {
+        'id': 'aionlabs',
+        'label': 'Aion',
+    },
+    {
+        'id': 'aoru',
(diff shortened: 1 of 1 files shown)


platberlitz (approved):
Thanks for the contribution!

No merge blockers found and nothing of note to fix. Merging.

PR #837, opened 2026-09-08
fix: Device setting persistence per-device

TheLonelyDevil9:
Does what it says. Mobile settings stay set for mobile, and other platforms like PC get their own persistence.


platberlitz:
No regressions, merging.

PR #840, opened 2026-09-08
fix: touch scrolling on mobile for blown-up text editors

TheLonelyDevil9:
Character definitions and other full-size text editors had no touch targets. Scrolling works as intended with this PR now.


platberlitz:
No regressions, merging.

PR #842, opened 2026-09-10
feat: Clipboard copy when copying screenshot of message(s)

TheLonelyDevil9:
Basically, whenever you click 'screenshot message(s)' in the chat UI, SB also copies the image to the clipboard for faster copy-pasting in various platforms ala discord.

### DIFF
--- a/public/script.js
+++ b/public/script.js
@@ -12812,27 +12812,62 @@ async function renderMessageScreenshotCanvas(startId, endId) {
     }
 }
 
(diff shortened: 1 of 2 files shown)


Geechan:
@TheLonelyDevil9 I would change the wording to "Screenshot messages to clipboard" to make it evident that this is the intended behaviour. Otherwise, a user could have their clipboard cleared inadvertently without enough indication that this will happen.

TheLonelyDevil9:
Addressing and merging

PR #845, opened 2026-09-15
fix: Upstream alignment for fixes/implementation

TheLonelyDevil9:
As discussed, implementing how SillyTavern does it for uniformity as we prepare for v1.7.1.

Here are the fixes/changes (by Astra):

- **Fable 5.1 structured replies:** Uses the model's built-in support for replies in a required JSON format while keeping the selected thinking effort. Other Claude models keep their existing behavior.
- **Image captions:** Adds Fable 5.1 to the caption model selector.
- **Astra request settings:** Sends the correct reply-length setting and removes unsupported options across OpenAI, Azure, OpenRouter, and SillyBunny's Responses connection. Profile overrides and requests from external apps follow the same rules.
- **Astra tool use:** Recognizes Astra variants and blocks unsupported tool calls through OpenAI and Azure Chat Completions. SillyBunny's Responses connection keeps its existing tool restriction.
- **Astra reply detail:** Preserves the selected verbosity, which controls how detailed replies should be, for OpenAI Chat Completions and Responses.
- **Astra token counts:** Uses the same token counter as SillyTavern for ElectronHub and server-side requests, including model names with a provider prefix.
- **Context limits:** Adopts SillyTavern's limits for how much text each model can handle, including Astra's 1.05 million tokens. SillyBunny's existing context-unlock behavior is preserved.
- **Expression lists:** Restores `{{availableExpressions}}` to return the full list of recognized and custom expressions. The companion keeps using the separate `{{availableSprites}}` list based on available character images.
- **Failed character renames on Windows:** Uses SillyTavern's file-copy fix when restoring chats after a character rename fails partway through.

### Testing

- All 108 tests in the six focused suites passed, along with lint, frontend build and budget checks, and Node/Bun checks using mock providers.
- The full suite had 3,702 passing tests and 39 failures. All 39 failures also occurred on the unchanged base and concern existing Windows filesystem and server-plugin tests.
- Live provider calls and browser end-to-end tests were not run.


Geechan (approved):
Looks good and I see no issues after reviewing the code.

PR #846, opened 2026-09-15
chore: load extra tools only when needed

TheLonelyDevil9:
SB now waits until you use some of the tools, to load them, lazy-loading them, to further boost time-to-open.

Server, Console Logs and Data Maid load when opened.

Some settings controls and suggestions are also prepared only when needed. 

Search still finds settings in panels you have not opened yet. If a tool fails to load, you can retry, and switching to another panel while it loads, it keeps your latest choice.

Testing: 59 focused automated tests and 8 browser tests passed. The app build and code-quality checks also passed. Tests confirm that startup loads less code and does less setup work, although the change in loading time varied between runs. The full GitHub test run also passed.


TheLonelyDevil9:
Comments are now human (except testing notes)

Geechan (approved):

PR #847, opened 2026-09-15
chore: make replies and chat history smoother

TheLonelyDevil9:
This PR makes less repeat work happen while a reply arrives. When multiple parts of a reply arrive together, it prepares the newest text for display, and keeps the older parts that have already been drawn.

Also, long chats load in smaller chunks so that SB gets chances to respond to clicks and scrolling.

If you delete a message or switch chats while loading, it checks the current chat before continuing.

This prevents duplicated messages, controls acting on the wrong message, and missing Companion content. 

Formatting, links, code blocks and the Stop button are covered by the checks as well.

Testing: 269 focused automated tests and 5 browser tests passed, and all GitHub code checks passed. Separate desktop and phone-sized browser checks confirmed correct final text and Stop behavior. In a test where many pieces of a reply arrived together, the app prepared the text once instead of 49 times. These improvements cover displaying replies; the AI service still determines how quickly new text is generated. A single very large message can still take time to display.


TheLonelyDevil9:
Comments are now human (except testing notes)

Geechan (approved):
No regressions or scope creep spotted, and no further issues. Safe to merge.

PR #848, opened 2026-09-15
chore: reconcile main release history with staging

TheLonelyDevil9:
Reconcile main's release-merge ancestry with staging so the branch comparison has no commits behind main. The sole main-only commit is 3021edcae, the 1.7.0 release merge from #739; its file tree matches its staging parent.

This branch merges main into staging and changes no files. Merge this PR using a merge commit so main remains an ancestor; squash or rebase merging would lose the ancestry relationship needed to clear the behind count.

Verified that the branch's tree matches staging exactly and that main is an ancestor of the branch. After merging, compare staging against main and confirm the behind count is zero.


TheLonelyDevil9:
Since no files change, merging

Issue #368, opened 2026-06-08
[MAJOR BUG] Toggling a Lot of Prompts/Agents can IRRECOVERABLY CORRUPT Your Chats


TRUEDarkAngel:
I was writing something different but now I must write this. I originally was going to write for help with the Shard Memory QR where even with all regex and all extensions disabled, it would always stop at '[DEVELOPMENTS]', have it say that it's still streaming, and return nothing. While I was going to seek help, I discovered a MASSIVE bug.

I was going to test if it still happened with all my tracker agents disabled. I turned them all off rather quickly and the entire chat was WIPED. The entire 600+ message chat was ERASED with a few instances of `Found an invalid or corrupted chat file` in the console. Thankfully the `backups` folder does exist and I was able to recover it, but this is still MASSIVE and should be fixed ASAP.

I copied the chat and tried it again and, while it did present the integrity check thing, it didn't save the chat file.

platberlitz:
Will work on it ASAP.

TRUEDarkAngel:
The corruption can also happen if you ENABLE agents quickly. Maybe it only applies to the tracker agents, I dunno. It happens when you toggle while the chat is refreshing due to the disabled regex I think maybe.

(Still no clue what the issue with the QR is, though, I think it might be related to the tracker agents? No clue.)

platberlitz:
QR is likely not related to the agents, but is part of the Extension itself.

TRUEDarkAngel:
(Just that QR worked like once when I disabled things, the issue is more so that while the tokens/response are still generating, the chat file just isn't updating. Less of a priority, rn, just wanted to bring it up since that was the original thing I was doing.)

TRUEDarkAngel:
MAJOR:

This issue can ALSO happen if an agent with a regex is toggled while a message is being edited.
@platberlitz

TheLonelyDevil9:
Solved definitively in PR 370.

PR #849, opened 2026-09-16
fix(chats): stage chat reloads and enforce lifecycle barriers to prevent chat wipes (#368)

cspiritsong:
note: fixes #368 

okay, so the thing is if you click reload, fiddle with extensions or whatever instead of waiting for the gen to finish, your chat session can get screwed. That was the problem, if I understood correctly. 

So the fix I'm proposing with this patch is something like an emergency shutdown. 

it basically forces the screen to remain there (so there will be visual perception of lag or whatever), but in reality its forcing the stream to 'quickly, finish streaming!!!!' before saving the state. 

Or, it will forcibly stop / cancel the stream and then quickly saves pending edits before reloading. 

the idea is if shit happens, it should at least keep the screen intact instead of wiping the chat to zero with no way to recover (hopefully)


TheLonelyDevil9:
Summary of fix as per Astra:

```Fixed group reloads to fetch the active chat-file ID rather than the group’s metadata ID, so branches and alternate group chats reload the correct file.

Reload now stops if saving the interrupted generation returns false or throws. It also rechecks chat identity after waiting for generation to stop, preventing navigation during that wait from saving against the wrong target.

Replaced the simulated differential tests with tests that execute the actual reload function. Coverage includes group branches, failed saves, successful partial-output persistence, failed fetches, and navigation during reload.
```

TheLonelyDevil9:
Good to go with the fixes. Merging.

PR #850, opened 2026-09-16
Fix: Initialize new group chats and restore group highlighting

voldomero:
Fixes two issues when creating a group:

- `src/endpoints/groups.js`: Creates the initial chat-file header before saving group metadata, preventing the “Could not load chat data” error on first open. Existing chat files are preserved.
- `public/script.js`: Corrects the highlight selector to use `data-grid`, skips entries that disappear during navigation, and catches asynchronous failures.


voldomero:
Just noticed this is in the wrong branch. My apologies.

PR #851, opened 2026-09-16
fix: Initialize new group chats, restore group highlighting and fix search in popout.

voldomero:
Fixes three issues found regarding Group Chats:

- `src/endpoints/groups.js`: Creates the initial chat-file header before saving group metadata, preventing the “Could not load chat data” error on first open. Existing chat files are preserved.
- `public/script.js`: Corrects the highlight selector to use `data-grid`, skips entries that disappear during navigation, and catches asynchronous failures.
- `public/scripts/group-chat.js`: Fixes search in the Current Members popout. The copied search field wasn’t connected to the filtering function, so typing didn’t update the member list.

_All issues reproduced with current project for group chats fully disabled._
Verified filtering, clearing, reopening, switching groups, fuzzy search, and mobile behavior.


TheLonelyDevil9:
Looks good to go. Merging.

PR #853, opened 2026-09-16
feat(samplers): add unified 3-state transmission policy and model-scoped wire sanitization

cspiritsong:
Okay, so different LLMs have different samplers and parameters. Most are hardcoded. So the problem is if users add 'exclude parameters', this affects ALL connections and saves into presets. So those with multiple connections like Grok, Gemini, GPT, will always need to remember which LLM hates / don't like what, and... trouble. Very off putting.

My proposed patch is to get rid of this pain point, but still allow the flexibility of whatever third party providers may allow to have their parameters turned on and off at will.

So the idea is:

I added a 3-button control directly under each sampler card in the sidebar:
- **Auto:** Default backend behavior.
- **Send:** Forces the slider's exact value to be sent. (User override)
- **Omit:** Strictly strips the key from the request JSON (so picky APIs don't 400, or in short, tell you "shit I can't do this" and ragequit on users).

The states are saved per model target, so it leaves presets completely untouched and won't mess up your other connections. Sliders also dim when omitted so it's visually obvious.

77 unit tests added and passing; existing preset and sampling tests pass cleanly.


TheLonelyDevil9:
Summary of fixes as per Astra:

```
Fixed Text Completions controls to read and write policies for the active text backend/model rather than the selected Chat Completions target.

Corrected Typical P handling to operate on the actual flat request fields, including the legacy typical alias. Omit now removes both fields, and Send keeps their values consistent. Also fixed stale custom-profile selections affecting native-provider policy keys, and prevented settings loading from overwriting normalized policy data.

Policy buttons and provider-restriction badges now refresh when the target changes. Removed unnecessary CSS !important overrides and conflicting inline styling, restoring the existing CSS budget.

Updated the integration-test harnesses to use the real policy implementation and added request-level coverage for omission, inclusion, and target isolation. Verified the controls in desktop and mobile browser layouts.
```

TheLonelyDevil9:
Good to go with the fixes. Merging.

